PR #8, opened 2021-11-04
Updated Notebook Containers for Tensorflow and Pytorch

mbaijal:
Updated Notebook Containers to Tensorflow 2.6 on and Pytorch 1.9 both on python3.8.

### DIFF
--- a/apps/jupyter/jupyter-web-app/upstream/base/configs/spawner_ui_config.yaml
+++ b/apps/jupyter/jupyter-web-app/upstream/base/configs/spawner_ui_config.yaml
@@ -20,11 +20,11 @@ spawnerFormDefaults:
     value: public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-scipy:v1.3.1-rc.0
     # The list of available standard container Images
     options:
+    - public.ecr.aws/c9e4w0g3/notebook-servers/jupyter-tensorflow:2.6.0-gpu-py38-cu112
+    - public.ecr.aws/c9e4w0g3/notebook-servers/jupyter-tensorflow:2.6.0-cpu-py38
+    - public.ecr.aws/c9e4w0g3/notebook-servers/jupyter-pytorch:1.9.0-gpu-py38-cu111
+    - public.ecr.aws/c9e4w0g3/notebook-servers/jupyter-pytorch:1.9.0-cpu-py38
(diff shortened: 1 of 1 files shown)


goswamig (approved):

apps/jupyter/jupyter-web-app/upstream/base/configs/spawner_ui_config.yaml
```diff
@@ -20,11 +20,10 @@ spawnerFormDefaults:
     value: public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-scipy:v1.3.1-rc.0
     # The list of available standard container Images
     options:
-    - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-scipy:v1.3.1-rc.0
-    - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-full:v1.3.1-rc.0
-    - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-cuda-full:v1.3.1-rc.0
-    - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-full:v1.3.1-rc.0
-    - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-cuda-full:v1.3.1-rc.0
+    - 984081543289.dkr.ecr.us-west-2.amazonaws.com/tensorflow-2.6-notebook-gpu:latest
```
surajkota, line 23: 1. see the images above, we should use ECR public as well
2. Please use versions in the tags instead of latest, lets follow DLC tagging scheme, just change the repository name
3. Should we just keep the default images at end of the list? definitely scipy since we don't have that, for others, do we have all packages that other images have, wondering if they will be useful for running some examples KF provides

surajkota (changes requested):
few questions/comments

apps/jupyter/jupyter-web-app/upstream/base/configs/spawner_ui_config.yaml
mbaijal: 1. The idea was to use private repos like DLC but as discussed I am changing them to public now. 
2. Yes, repo names and tagging needed to be reviewed, editing. 
3. We have all the packages that are in the default images and newer versions, plus the default images seem to be broken so I do want to remove them. Can add back the scipy image till I find the appropriate replacement.

goswamig (approved):

PR #13, opened 2021-11-07
Fixing pipeline and Katib steps

goswamig:
Resolves #

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

Manually tested pipeline and make sure it has worked.

### DIFF
--- a/examples/aws/cognito-rds-s3/README.md
+++ b/examples/aws/cognito-rds-s3/README.md
@@ -56,9 +56,6 @@ Follow the [Configure Katib](../rds-s3/README.md#3-configure-katib) section from
         # Kubeflow Istio Resources
         kustomize build common/istio-1-9/kubeflow-istio-resources/base | kubectl apply -f -
         
-        # Kubeflow Pipelines
-        # reapply manifest if you see an error
-        kustomize build apps/pipeline/upstream/env/aws | kubectl apply -f -
         
         # KFServing
         kustomize build apps/kfserving/upstream/overlays/kubeflow | kubectl apply -f -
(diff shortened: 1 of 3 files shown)


examples/aws/cognito-rds-s3/README.md
```diff
@@ -111,16 +111,10 @@ Follow the [Configure Katib](../rds-s3/README.md#3-configure-katib) section from
 
         # Envoy filter
         kustomize build distributions/aws/aws-istio-envoy-filter/base | kubectl apply -f -
-        
-        # Configured for AWS RDS and AWS S3
```
surajkota, line 115: nit: afaik, @rrrkharse mentioned something related to ordering because of which he kept these steps at the end
also for uniformity, let's keep modified components after vanilla

can you move this section above Ingress?

generic components
RDS,S3
Cognito

Same for here - https://github.com/awslabs/kubeflow-manifests/blob/v1.3-branch/examples/aws/cognito-rds-s3/kustomization.yaml. I just noticed the ordering in this file is also reverse

surajkota (changes requested):
Good catch. Thanks for bug fixes!

examples/aws/cognito/README.md
```diff
@@ -97,7 +97,7 @@ Follow this step only for `*.platform.example.com`:
     1. The Pool ARN of the user pool found in Cognito general settings.
     2. The App client id, found in Cognito App clients.
     3. The custom user pool domain (e.g. `auth.platform.example.com`), found in the Cognito domain name.
-    4. The ARN of the certificate from the Certificate Manager in the region where your platform is running.
+    4. The ARN of the certificate from the Certificate Manager in the region where your platform (subdomain) is running.
```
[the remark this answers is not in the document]
goswamig: done

examples/aws/cognito-rds-s3/README.md
goswamig: Done first two parts. 

I like the idea of having the order consistent. But I am afraid on making changes which might break at the last min. we know authentication works in RDS_s3 side so let it be for now...we can fix the order later.

surajkota (approved):

PR #55, opened 2021-12-21
Secret manager integration with katib

ryansteakley:
 

- Utilize AWS secret manager with Katib
- Update readme to create rds secret with necessary secrets for Katib Integration
   - Can possibly just have 1 command to create the rds-secret and mention that it will be for rds and katib
   
**Testing**
Created an experiment and verified result on web-ui
Screen Shot 2021-12-20 at 4 40 08 PM
Screen Shot 2021-12-20 at 4 39 59 PM

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


distributions/aws/examples/rds-s3/README.md
```diff
@@ -118,15 +118,17 @@ kubectl apply -f https://raw.githubusercontent.com/aws/secrets-store-csi-driver-
...
-      KATIB_MYSQL_DB_PORT=3306
-      DB_USER=admin
-      DB_PASSWORD=Kubefl0w
+      dname=kubeflow
+      host=rm12abc4krxxxxx.xxxxxxxxxxxx.us-west-2.rds.amazonaws.com
+      port=3306
+      username=admin
+      password=Kubefl0w
+
+      aws secretsmanager update-secret  --region $CLUSTER_REGION --secret-id rds-secret --secret-string '{"username":"admin","password":"Kubefl0w","dbname":"kubeflow","host":"rm12abc4krxxxxx.xxxxxxxxxxxx.us-west-2.rds.amazonaws.com","port":"3306"}'
```
rrrkharse, line 131: does the current implementation support using a separate database for katib and KFP?

currently it seems like since both are sharing the same secrets that both will have to use the same RDS instance, but customers may want to scale these seperately

ryansteakley: Good point if that’s the case I can separate the secrets then

apps/pipeline/upstream/env/aws/secrets-manager.yaml
```diff
@@ -13,6 +13,12 @@ spec:
       key: username
     - objectName: "pass"
       key: password
+    - objectName: "host"
```
surajkota, line 16: move secrets manager related files to a common dir?

remember to update the steps in readmes with any changes required

distributions/aws/examples/rds-s3/README.md
surajkota: That was the case before as well, we are asking user to create one RDS instance in pre-requisites. Lets not change the approach in this PR since it is a 2 way door.

@ryansteakley lets create the secret in the pre-requisites itself. Previously, user had to substitute vars in multiple files and hence we had separate instructions for pipelines and Katib. Now, it can be a single step

Also, can you check if it is possible to configure the rds endpoint for pipelines from the secret as well?

surajkota (changes requested):

distributions/aws/examples/rds-s3/README.md
surajkota: remove the unused `.env` files

apps/pipeline/upstream/env/aws/secrets-manager.yaml
goswamig: Which all files are we talking here ? Does the common dir lives under aws?

distributions/aws/examples/rds-s3/README.md
goswamig: folks two things here.
1. I think its fine for now to keep the same secret (which will ultimate result into same RDS) for both pipeline and Katib.
2. However lets have instruction separete for katib and pipeline in such a way that customer can pick and use one of them or combination. 
3. Why do we want RDS endpoint to be part of secret ?

ryansteakley: 1. I don't think it is necessary to have separate instructions for katib and the pipelines since the one secret will be used for both. If they choose to use one of them or in combination the steps wouldn't be different.
2. It should be possible to configure the rds endpoint for pipelines using secrets since it is using a configmap and creating env vars from there but can change to pull from a secret instead.

distributions/aws/examples/rds-s3/README.md
```diff
@@ -102,33 +118,6 @@ kubectl apply -f https://raw.githubusercontent.com/aws/secrets-store-csi-driver-
...
-          aws secretsmanager create-secret --name rds-secret --secret-string '{"username":"admin","password":"Kubefl0w"}'  --region $CLUSTER_REGION
-          ```
-    1. Configure an AWS secret with your AWS credentials. These need to be long term credentials from an IAM user and not temporary. 
-        - Find more details about configuring/getting your AWS credentials here:
-        https://docs.aws.amazon.com/general/latest/gr/aws-security-credentials.html
-        - ```
-          aws secretsmanager create-secret --name s3-secret --secret-string '{"accesskey":"AXXXXXXXXXXXXXXXXXX6","secretkey":"eXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXq"}'  --region $CLUSTER_REGION
-          ```
-
-### 3. Configure Katib
```
surajkota, line 118: you will need to update the cognito-rds-s3 readme according to the changes in this readme like configure Katib section does not exist anymore

mainly the first 3 sections, Prerequisites, Configure Kubeflow Pipelines for RDS and S3 and Configure Katib for RDS

distributions/aws/examples/rds-s3/README.md
```diff
@@ -63,35 +63,51 @@ aws s3 mb s3://$S3_BUCKET --region $CLUSTER_REGION
 
 Follow this [doc](https://www.kubeflow.org/docs/distributions/aws/customizing-aws/rds/#deploy-amazon-rds-mysql) to set up an AWS RDS instance. Please follow only section called `Deploy Amazon RDS MySQL`. 
 
-6. Install AWS Secrets & Configuration Provider with Kubernetes Secrets Store CSI driver
+6. Create Secrets
+    1. Configure the secret with the RDS DB name, RDS endpoint URL, RDS DB port, and RDS DB credentials that were configured when following the steps in Create RDS Instance.
+        - For example, if your database name is `kubeflow`, your endpoint URL is `rm12abc4krxxxxx.xxxxxxxxxxxx.us-west-2.rds.amazonaws.com`, your DB port is `3306`, your DB username is `admin`, and your DB password is `Kubefl0w` your secret should look like:
+        - **Note:** These are the default values for the database name and credentials in cloudformation template for creating the RDS instance, change these according to the values you used
+        - ```
+          dname=kubeflow
```
surajkota, line 71: this block of values before the actual command is confusing. I think it can be safely removed

distributions/aws/examples/rds-s3/README.md
```diff
@@ -63,35 +63,51 @@ aws s3 mb s3://$S3_BUCKET --region $CLUSTER_REGION
 
 Follow this [doc](https://www.kubeflow.org/docs/distributions/aws/customizing-aws/rds/#deploy-amazon-rds-mysql) to set up an AWS RDS instance. Please follow only section called `Deploy Amazon RDS MySQL`. 
 
-6. Install AWS Secrets & Configuration Provider with Kubernetes Secrets Store CSI driver
+6. Create Secrets
+    1. Configure the secret with the RDS DB name, RDS endpoint URL, RDS DB port, and RDS DB credentials that were configured when following the steps in Create RDS Instance.
```
surajkota, line 67: int: Configure a secret named `rds-secret`, same for s3-secret

distributions/aws/examples/rds-s3/README.md
```diff
@@ -63,35 +63,51 @@ aws s3 mb s3://$S3_BUCKET --region $CLUSTER_REGION
 
 Follow this [doc](https://www.kubeflow.org/docs/distributions/aws/customizing-aws/rds/#deploy-amazon-rds-mysql) to set up an AWS RDS instance. Please follow only section called `Deploy Amazon RDS MySQL`.
```
surajkota, line 64: Can you add a line here saying this database instance will be used by Pipelines and Katib
and add a similar text in S3 step saying this bucket will be used to store artifacts from pipelines

asking because we are combing a few things now(e.g. no section for Katib) and it can be confusing for a user who does not know where/how these database and S3 bucket used in Kubeflow

distributions/aws/examples/rds-s3/README.md
```diff
@@ -63,35 +63,51 @@ aws s3 mb s3://$S3_BUCKET --region $CLUSTER_REGION
 
 Follow this [doc](https://www.kubeflow.org/docs/distributions/aws/customizing-aws/rds/#deploy-amazon-rds-mysql) to set up an AWS RDS instance. Please follow only section called `Deploy Amazon RDS MySQL`. 
 
-6. Install AWS Secrets & Configuration Provider with Kubernetes Secrets Store CSI driver
+6. Create Secrets
```
surajkota, line 66: nit: Create Secrets in AWS Secrets Manager?

surajkota (changes requested):

distributions/aws/examples/cognito-rds-s3/README.md
```diff
@@ -8,16 +8,9 @@ Follow the pre-requisites section from [this guide](../rds-s3/README.md#1-prereq
 1. Clone the repo
 1. Create an EKS cluster and
 1. Create S3 Bucket
-
-## Configure Kubeflow Pipelines for RDS and S3
```
surajkota, line 12: user still need to follow this in addition to prereqs. maybe you can change the section name to
`Configure RDS endpoint and S3 bucket name for Kubeflow Pipelines` to make it clear

distributions/aws/examples/rds-s3/README.md
```diff
@@ -61,37 +62,47 @@ aws s3 mb s3://$S3_BUCKET --region $CLUSTER_REGION
...
 
-6. Install AWS Secrets & Configuration Provider with Kubernetes Secrets Store CSI driver
+6. Create Secrets in AWS Secrets Manager
+    1. Configure a secret named rds-secret with the RDS DB name, RDS endpoint URL, RDS DB port, and RDS DB credentials that were configured when following the steps in Create RDS Instance.
+        - For example, if your database name is `kubeflow`, your endpoint URL is `rm12abc4krxxxxx.xxxxxxxxxxxx.us-west-2.rds.amazonaws.com`, your DB port is `3306`, your DB username is `admin`, and your DB password is `Kubefl0w` your secret should look like:
+        - **Note:** These are the default values for the database name and credentials in cloudformation template for creating the RDS instance, change these according to the values you used
+        - ```
+          aws secretsmanager create-secret --name rds-secret --secret-string '{"username":"admin","password":"Kubefl0w","dbname":"kubeflow","host":"rm12abc4krxxxxx.xxxxxxxxxxxx.us-west-2.rds.amazonaws.com","port":"3306"}' --region $CLUSTER_REGION 
+          ```
+    1. Configure a secret named s3-secret with your AWS credentials. These need to be long term credentials from an IAM user and not temporary.
```
surajkota, line 74: super nit: `s3-secret` same for rds

surajkota (changes requested):

surajkota (approved):
Lets :ship: it

PR #61, opened 2021-12-30
cognito e2e test

surajkota:
- Moved cognito bootstrap scripts so they are reusable for tests + it will be tested and up to date
- Added tests for cognito deployment
- Some utils and helper methods
- Issue with using kfp_client programmatically with cognito setup. Described the limitation in the test.

**Testing:**
- Tested using metadata
```
pytest -s --keepsuccess --region ca-central-1 --root-domain-name xxxx.dev --root-domain-hosted-zone-id XXXXXX --metadata .metadata/metadata-1640755255480551966.json tests/test_cognito_deployment.py
```
- Tested without using metadata and keep success flags as well and made changes to clean up scripts for leaking resources. Load balancer was not getting deleted because the alb-controller gets deleted almost same time as ingress.
- Tested cognito-automated instructions after the path changes


rrrkharse (approved):
Looks good, just some nits.

Also was wondering if you have run it without the `--metadata` and `--keepsuccess` flags (e.g. from scratch e2e). I will make it a point to include the output without these flags as well because sometimes reusing the metadata hides test failures that occur when no metadata is present.

distributions/aws/test/e2e/fixtures/cognito_dependencies.py
```diff
@@ -0,0 +1,249 @@
...
+    role_name = rand_name("kf-admin-role")
+    iam_client = get_iam_client(region)
+    iam_resource = boto3.resource("iam")
+
+    def on_create():
+        associate_iam_oidc_provider(cluster, region)
+        eks_client = get_eks_client(region)
+        response = eks_client.describe_cluster(name=cluster)
+
+        oidc_provider_url = response["cluster"]["identity"]["oidc"]["issuer"][8:]
```
ryansteakley, line 110: will this string split always be consistent at 8 chars in future?

surajkota: yes, we need to remove `https://` it has not changed since 2019 atleast

distributions/aws/test/e2e/tests/test_cognito_deployment.py
```diff
@@ -0,0 +1,53 @@
...
+)
+
+
+@pytest.fixture(scope="class")
+def kustomize_path():
+    return "../../examples/cognito"
+
+
+@pytest.fixture(scope="class")
+def configure_manifests(region, cluster, configure_ingress, cognito_bootstrap):
```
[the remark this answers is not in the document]
surajkota: Good point. we dont want to keep these as required parameters. I will add a check

ryansteakley (approved):

distributions/aws/test/e2e/fixtures/kustomize.py
```diff
@@ -42,7 +42,7 @@ def delete_kustomize(path):
 
 
 @pytest.fixture(scope="class")
-def kustomize(metadata, cluster, kustomize_path, request):
+def kustomize(metadata, cluster, configure_manifests, kustomize_path, request):
```
mbaijal, line 45: Does this break @rrrkharse's test PR ?

distributions/aws/test/e2e/tests/test_cognito.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_url_is_up(self, setup, cognito_bootstrap):
+        subdomain_name = cognito_bootstrap["route53"]["subDomain"]["name"]
+        kubeflow_endpoint = "https://kubeflow." + subdomain_name
+        response = requests.get(kubeflow_endpoint)
+        assert response.status_code == 200
+        # request was redirected
+        assert len(response.history) > 0
+        # redirection was to cognito domain
+        assert "auth." + subdomain_name in response.url
```
mbaijal, line 46: This is not testing login correct but just url redirection. Are we planning to test login ?

mbaijal (approved):
Looks good !

distributions/aws/test/e2e/fixtures/kustomize.py
surajkota: No, rds-s3 pr has similar changes

distributions/aws/test/e2e/tests/test_cognito.py
surajkota: We can test login using a browser but not using API calls right now. Browser testing can be separate PR

another way would be to use kubeconfig instead of session or cookies but I am not sure if its the right UX and we are not testing multi-user at that point.

PR #64, opened 2022-01-06
Generate unit tests 

ryansteakley:
- Add directories of the kustomization packages we are modifiyng/own to `generate_test.py`
- Run these new generated tests in `tests/Makefile `
- Attempted to change the output of `generate_test.py` to our `distributions/aws` however this caused issues in the tests being unable to run as they are missing the required files needed to run
   - Even if adding these required files the tests do not run do to path failures/ seems to be an issue of tests being too nested, same issue can be seen with old unit tests in the `tests` directory not running for the same issue if they are one folder deep.
- In future can look into additional changes to host these tests in distributions but as of now I think having these tests in the `tests` directory is acceptable.
- Non-auto generated files
  - `hack/generate_tests.py`
  - `tests/Makefile`
**Checklist:**
- [ X] Unit tests pass:
    2. `make test`


rrrkharse (approved):

tests/apps/jupyter/jupyter-web-app/upstream/base/test_data/expected/apps_v1_deployment_jupyter-web-app-deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - env:
+        - name: APP_PREFIX
+          value: /jupyter
+        - name: UI
+          value: default
+        - name: USERID_HEADER
+          value: kubeflow-userid
+        - name: USERID_PREFIX
+          value: ""
+        image: public.ecr.aws/j1r0q0g6/notebooks/jupyter-web-app:v1.3.1-rc.0
```
mbaijal, line 33: future nit: we should look into why an `rc` image is being used here

mbaijal (approved):

PR #70, opened 2022-01-18
AWS sanity canary test files 

akartsky:

Add files to run sanity canary tests

### DIFF
--- a/distributions/aws/test/canary/scripts/run_test.sh
+++ b/distributions/aws/test/canary/scripts/run_test.sh
@@ -29,6 +29,6 @@ mkdir -p $E2E_TEST_DIR/.metadata/
 cp metadata-canary $E2E_TEST_DIR/.metadata/
 
 cd $E2E_TEST_DIR
-pytest -s -q --metadata .metadata/metadata-canary --keepsuccess --region $CLUSTER_REGION
+pytest tests/test_sanity.py -s -q --metadata .metadata/metadata-canary --keepsuccess --region $CLUSTER_REGION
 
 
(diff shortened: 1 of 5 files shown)


ryansteakley (approved):

distributions/aws/test/canary/scripts/run_test.sh
```diff
@@ -0,0 +1,34 @@
...
+# Connect to eks cluster 
+aws eks update-kubeconfig --name $CLUSTER_NAME --region $CLUSTER_REGION
+
+# Modify metadeta file
+cd $CANARY_TEST_DIR
+sed -i 's/$CLUSTER_NAME/'"$CLUSTER_NAME"'/g' metadata-canary
+sed -i 's/$CLUSTER_REGION/'"$CLUSTER_REGION"'/g' metadata-canary
+
+mkdir -p $E2E_TEST_DIR/.metadata/
+cp metadata-canary $E2E_TEST_DIR/.metadata/
```
mbaijal, line 29: Can we not copy the `metadata-canary` file over first and then edit so we don't edit the source ?

mbaijal (approved):

distributions/aws/test/canary/scripts/run_test.sh
surajkota: @akartsky instead of maintaining a separate file and using sed command for each variable, what do you think about using `printf` command directly to write the contents to the file

```
printf '{
"region": '$CLUSTER_REGION'
}' > .metadata/canary-meta.json
```

akartsky: @mbaijal 
It's happening inside the docker image so doesn't matter if we edit first or if we copy over and then edit 

@surajkota 
By having a dummy metadata file it's easier to see the skeleton of the file and even reuse it for local testing for example

surajkota: approving since this can be changed later if there are too many vars

surajkota (approved):

Issue #63, opened 2022-01-06
AWS ALB Ingress controller runs into "Reconciler error"


mmuppidi:
I am trying to install Kubeflow 1.3 on an EKS 1.19 cluster using the instructions provided [here](https://github.com/awslabs/kubeflow-manifests/tree/v1.3-branch/distributions/aws/examples/cognito). We have a shared VPC setup with 2 private subnets, each subnet has appropriate tags and deployed in different AZs

| Key | Value|
| --- | ----------- |
| kubernetes.io/cluster/kubeflow-dev | shared |
| kubernetes.io/role/internal-elb | 1 |

I configured the IAM Role backed service account as described in the instructions as well. I also checked that the IAM role has appropriate permissions attached to it. When the AWS Ingress controller comes up it continuously runs into the following error

```bash
E0106 03:07:35.678637       1 controller.go:217] kubebuilder/controller "msg"="Reconciler error" "error"="failed to build LoadBalancer configuration due to retrieval of subnets failed to resolve 2 qualified subnets. Subnets must contain the kubernetes.io/cluster/\u003ccluster name\u003e tag with a value of shared or owned and the kubernetes.io/role/elb tag signifying it should be used for ALBs Additionally, there must be at least 2 subnets with unique availability zones as required by ALBs. Either tag subnets to meet this requirement or use the subnets annotation on the ingress resource to explicitly call out what subnets to use for ALB creation. The subnets that did resolve were []"  "controller"="alb-ingress-controller" "request"={"Namespace":"istio-system","Name":"istio-ingress"}
E0106 03:07:36.717626       1 controller.go:217] kubebuilder/controller "msg"="Reconciler error" "error"="failed to build LoadBalancer configuration due to retrieval of subnets failed to resolve 2 qualified subnets. Subnets must contain the kubernetes.io/cluster/\u003ccluster name\u003e tag with a value of shared or owned and the kubernetes.io/role/elb tag signifying it should be used for ALBs Additionally, there must be at least 2 subnets with unique availability zones as required by ALBs. Either tag subnets to meet this requirement or use the subnets annotation on the ingress resource to explicitly call out what subnets to use for ALB creation. The subnets that did resolve were []"  "controller"="alb-ingress-controller" "request"={"Namespace":"istio-system","Name":"istio-ingress"}
[...]

PR #80, opened 2022-02-04
Remove loadBalancerScheme from Cognito params

akartsky:
Customer can just use the loadBalancerScheme param in the base : https://github.com/awslabs/kubeflow-manifests/blob/v1.3-branch/distributions/aws/istio-ingress/base/params.env

If we do want to keep the parameter in Cognito folder then for it to work we have to change the parameter name `loadBalancerScheme` and edit other files in the folder.

### DIFF
--- a/distributions/aws/istio-ingress/overlays/cognito/params.env
+++ b/distributions/aws/istio-ingress/overlays/cognito/params.env
@@ -2,4 +2,3 @@ CognitoUserPoolArn=
 CognitoAppClientId=
 CognitoUserPoolDomain=
 certArn=
-loadBalancerScheme=internet-facing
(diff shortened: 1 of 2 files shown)


distributions/aws/istio-ingress/overlays/cognito/params.env
```diff
@@ -2,4 +2,3 @@ CognitoUserPoolArn=
 CognitoAppClientId=
 CognitoUserPoolDomain=
 certArn=
-loadBalancerScheme=internet-facing
```
surajkota, line 5: can you remove from cognito readme steps before merging? step 3 of https://github.com/awslabs/kubeflow-manifests/tree/v1.3-branch/distributions/aws/examples/cognito#40-configure-ingress

surajkota (approved):

surajkota:
#63

PR #96, opened 2022-02-19
[WIP] Enhance EFS Tests: Dynamic Provisioning and validate using KFP

mbaijal:
#### [WIP] This PR is complete in functionality but needs some code cleanup

 - [x] Adds a test for EFS dynamic provisioning using the automated script, `auto-efs-setup.py`
 - [x] added a parameter to the auto-install script to specify a path to the `sc.yaml` file since the path requirements change when running from tests - this can be customized a bit more so that I can use it for static provisioning as well. 
 - [x]  Enhance the EFS static and dynamic provisioning tests to also test the volume usage using 2 KFP pipelines - one writes to the volume and other reads from it. 
 
**TODO:**
- [ ] Static Provisioning also needs an automated script. 
- [ ] The KFP test is currently taking very long to run on my cluster - about 40minutes sometimes - this has increased with the number of runs on my cluster so possibly a resource limitation - I need to debug this.

**Testing:** - 

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/distributions/aws/examples/storage/efs/auto-efs-setup.py
+++ b/distributions/aws/examples/storage/efs/auto-efs-setup.py
@@ -5,7 +5,7 @@ import string
 import random
 import yaml
 import urllib.request
-
+import random
 from shutil import which
 from time import sleep
 
@@ -206,7 +206,8 @@ def kubectl_apply(file_name: str):
         "kubectl",
         "apply",
         "-f",
-        file_name
+        file_name, 
+        "--force"
     ])
 
 
@@ -537,6 +538,14 @@ def footer():
     print("                      EFS Setup Complete")
     print("=================================================================")
 
+def rand_name(prefix):
+    """
+    Returns a random string of 10 ascii lowercase characters appended to the prefix
+    """
+    suffix = "".join(
+        random.choice(string.ascii_lowercase + string.digits) for _ in range(10)
+    )
+    return prefix + suffix
 
 parser = argparse.ArgumentParser()
 parser.add_argument(
@@ -553,7 +562,7 @@ parser.add_argument(
     help='Your cluster name (eg: mycluster-1)',
     required=True
 )
-EFS_FILE_SYSTEM_NAME_DEFAULT = "KubeflowEfs"
+EFS_FILE_SYSTEM_NAME_DEFAULT = rand_name("Kubeflow-efs")
 parser.add_argument(
     '--efs_file_system_name',
     type=str,
--- a/distributions/aws/examples/storage/efs/auto-efs-setup.py
+++ b/distributions/aws/examples/storage/efs/auto-efs-setup.py
@@ -5,7 +5,6 @@ import string
 import random
 import yaml
 import urllib.request
-import random
 from shutil import which
 from time import sleep
 
@@ -44,12 +43,16 @@ def verify_oidc_provider_prerequisite():
     if is_oidc_provider_present():
         print("OIDC provider found")
     else:
-        raise Exception(f"Prerequisite not met : No OIDC provider found for cluster '{CLUSTER_NAME}'!")
+        raise Exception(
+            f"Prerequisite not met : No OIDC provider found for cluster '{CLUSTER_NAME}'!"
+        )
 
 
 def is_oidc_provider_present() -> bool:
     iam_client = get_iam_client()
-    oidc_providers = iam_client.list_open_id_connect_providers()["OpenIDConnectProviderList"]
+    oidc_providers = iam_client.list_open_id_connect_providers()[
+        "OpenIDConnectProviderList"
+    ]
 
     if len(oidc_providers) == 0:
         return False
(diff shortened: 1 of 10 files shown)


distributions/aws/test/e2e/tests/test_storage_efs.py
```diff
@@ -66,3 +71,43 @@ def test_pvc_with_volume(
...
+
+        fs_id = create_efs_volume["file_system_id"]
+        assert "fs-" in fs_id
+
+        efs_claim_dyn = dynamic_provisioning["efs_claim_dyn"]
+        claim_list = subprocess.check_output("kubectl get pvc -A".split()).decode()
+        assert efs_claim_dyn in claim_list
+
+        # TODO: Add this check to static provisioning as well
+        # TODO: Status is currently pending but we should also be testing for this to reach bound state
```
goswamig, line 110: Yes without bound state the test is incomplete.

distributions/aws/examples/storage/efs/auto-efs-setup.py
```diff
@@ -553,7 +562,7 @@ def footer():
     help='Your cluster name (eg: mycluster-1)',
     required=True
 )
-EFS_FILE_SYSTEM_NAME_DEFAULT = "KubeflowEfs"
+EFS_FILE_SYSTEM_NAME_DEFAULT = rand_name("Kubeflow-efs")
```
goswamig, line 565: Are we not deleting these EFS volume ?

AlexandreBrown, line 565: Could I get more context on this change as it seems to change the behaviour of the script?   

The default behaviour ensured that if the script was repeated it would not recreate resources (same behaviour for RDS & S3 script).  
The user would have to specify a new EFS file system name for a new file system to be created. The idea is that with a new file system might come a different configuration which is very hard to guess. For instance, will the user want to keep the same security group? I think It's harder to always be right when we take a stance like this.  
  
That being said, If we think always creating a new EFS file system and re-using the security group is probably the best default then that's fine as the user can also decide to pass a hardcoded EFS file name via the `--efs_file_system_name` parameter to prevent creating a new one.  
  
If we decide to go with this behaviour then we must take the following into account :  
- The RDS & S3 script behaviour would be different (the RDS & S3 script does not recreate a RDS and S3 bucket on every run of the script, to do so would require to generate a random RDS db instance name & random RDS secret name. It's doable but we just gotta keep that in mind.
- We should update the doc and make a clear mention of this behaviour so that the user has no surprise (unwanted new resources created).

distributions/aws/examples/storage/efs/auto-efs-setup.py
```diff
@@ -537,6 +538,14 @@ def footer():
     print("                      EFS Setup Complete")
     print("=================================================================")
 
+def rand_name(prefix):
```
surajkota, line 541: We should go the other way. Modify this script to use it in the tests

distributions/aws/test/e2e/tests/test_storage_efs.py
mbaijal: Yes that is a planned enhancement for this test already in the works.

distributions/aws/examples/storage/efs/auto-efs-setup.py
mbaijal: Not in this script no, so if you run this script multiple times it reuses the same volume. We need to randomize this name to be able to use these in the test automation.

mbaijal: More details are in the comment below by Alexandre.

mbaijal: That's a good point. Let me get back to you on this once I have the next change in place. 
I see that I could actually keep this script as before and instead pass a randomized EFS filename as a parameter so the test runs dont interfere with each other.

distributions/aws/examples/storage/efs/auto-efs-setup.py
mbaijal: Do you mean "don't modify this script but modify usage in tests instead" as I mentioned above ?

distributions/aws/test/e2e/tests/test_storage_efs.py
mbaijal: This change is now a part of the PR.

distributions/aws/examples/storage/efs/auto-efs-setup.py
mbaijal: I reverted the changes I made to your script and have used the script directly in the test scripts.

distributions/aws/examples/storage/efs/auto-efs-setup.py
mbaijal: done, if I understood you correctly, can you confirm.

mbaijal:
@AlexandreBrown Thanks for some very useful comments on this PR but I need to close it since the 1.3 release is now complete. I have ported the changes to the main branch however due to major directory structure refactoring, I have to create a new PR. Feel free to review the new PR #133 instead.

PR #103, opened 2022-02-25
Check cluster-region before creating s3-Bucket in automated setup

ryansteakley:
Automated setup will fail if cluster_region is set to us-east-1, since the s3-bucket creation will error out due to 

> botocore.exceptions.ClientError: An error occurred (InvalidLocationConstraint) when calling the CreateBucket operation: The specified location-constraint is not valid
thus will check if region is us-east-1 and call CreateBucket without this constraint.

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/distributions/aws/examples/rds-s3/auto-rds-s3-setup.py
+++ b/distributions/aws/examples/rds-s3/auto-rds-s3-setup.py
@@ -126,21 +126,12 @@ def does_bucket_exist(s3_client):
 def create_s3_bucket(s3_client):
     print("Creating S3 bucket...")
 
-    # CreateBucketConfiguration cannot have LocationConstraint of us-east-1 since it is the default
-    if CLUSTER_REGION == "us-east-1":
-        s3_client.create_bucket(
-        ACL="private",
-        Bucket=S3_BUCKET_NAME
-    )
-    else: 
-        s3_client.create_bucket(
-            ACL="private",
-            Bucket=S3_BUCKET_NAME,
-            CreateBucketConfiguration={
-                "LocationConstraint": CLUSTER_REGION,
-            }
-        )
-
+    args = {"ACL":"private","Bucket":S3_BUCKET_NAME}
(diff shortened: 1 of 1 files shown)


distributions/aws/examples/rds-s3/auto-rds-s3-setup.py
```diff
@@ -126,13 +126,20 @@ def does_bucket_exist(s3_client):
...
         ACL="private",
-        Bucket=S3_BUCKET_NAME,
-        CreateBucketConfiguration={
-            "LocationConstraint": CLUSTER_REGION,
-        }
+        Bucket=S3_BUCKET_NAME
     )
+    else: 
+        s3_client.create_bucket(
+            ACL="private",
```
surajkota, line 137: nit: one option to avoid duplication is by using a dict
```
args = {
ACL = "",

}
if us-east:
  args[configuration]={region}

create_bucket(**args)
```

surajkota (approved):

distributions/aws/examples/rds-s3/auto-rds-s3-setup.py
ryansteakley: good suggestion

AlexandreBrown:
Hello, could you explain why the if is needed ? Would it still work without the if when our cluster region is us-east-1 ?
Thanks

ryansteakley:
> Hello, could you explain why the if is needed ? Would it still work without the if when our cluster region is us-east-1 ? Thanks
@AlexandreBrown 
The if is needed since locationConstraint is a required field if you are using any region besides us-east-1. However if you are using us-east-1 then using locationConstraint will be invalid. The original code doesn't work if us-east-1 is our cluster region. and you get this error message 

> botocore.exceptions.ClientError: An error occurred (InvalidLocationConstraint) when calling the CreateBucket operation: The specified location-constraint is not valid
thus will check if region is us-east-1 and call CreateBucket without this constraint.

AlexandreBrown (approved):

PR #106, opened 2022-02-28
Added RDS & S3 Credentials Setup For Notebooks & Pipeline

### LINKED ISSUE #100: Facilitate configuration of RDS & S3 credentials access for Notebooks & Pipelines
It might not be obvious which k8s secret name to use to create a pod default so that we can use the RDS & S3 credentials setup by the credentials manager. It might not be ovious for new k8s users that you need to create something called a PodDefault in order to have access to these credentials from within a notebook and in notebook creation UI.  
  
**Use Cases**  
- Specify AWS credentials when using `boto3`   
  - Solution : Have a pod default make environment variables available 
- Use `ml-metadata` (requires database credentials in notebook)
  - Solution : Have a pod default make environment variables available

I propose we add this as part of the Kubeflow install.  
For instance, if the user installs Kubeflow via S3 setup, his pod default could look like 
```yaml
apiVersion: kubeflow.org/v1alpha1
kind: PodDefault
metadata:
  name: add-aws-secret
  namespace: <YOUR_NAMESPACE_HERE>
  labels:
    add-aws-secret : "true"
spec:
  desc: add-aws-secret
  selector:
    matchLabels:
      add-aws-secret: "true"
  env:
  - name: AWS_ACCESS_KEY_ID
    valueFrom:
      secretKeyRef:
        name: mlpipeline-minio-artifact
        key: accesskey
  - name: AWS_SECRET_ACCESS_KEY
    valueFrom:
      secretKeyRef:
        name: mlpipeline-minio-artifact
        key: secretkey
```
If he installs via RDS & S3 setup then we would add the RDS values (eg: DB_HOST, DB_PORT, DB_NAME, DB_USER, DB_PWD from `mysql-secret`) so that the user can [connect to ml metadata db](https://github.com/google/ml-metadata/blob/master/g3doc/get_started.md) in a notebook using OS environment variable values.  
  
This would also open the door to greater notebook integration as we could modify the `spawner_ui_config.yaml` and add the pod default created during the setup as default value selected.  
```yaml
  configurations:
    # List of labels to be selected, these are the labels from PodDefaults
    # value:
    #   - add-gcp-secret
    #   - default-editor
    value:
       - add-aws-secret
    readOnly: false
```
image

AlexandreBrown:
Resolves #100 

This PR adds instructions on how and when to setup RDS & S3 Credentials for Notebooks & Pipeline.  

https://user-images.githubusercontent.com/26939775/156860649-ffcdea02-6961-4ded-b23f-b83fb3d46bfe.mp4

 
**Checklist:**
- [x] Unit tests pass:
    2. `make test`
- [x] End-to-end test on fresh install (main branch + manifest v1.4.1)
- [x] Support S3 credentials
- [x] Support DB credentials


docs/deployment/rds-s3/README.md
```diff
@@ -248,7 +289,62 @@ mysql> select * from observation_logs;
...
+  configurations:
+    # List of labels to be selected, these are the labels from PodDefaults
+    value:
+      - add-aws-secret
+```  
+  
+This change can be done before or after installing Kubeflow, here is how to make this modification depending on these 2 options.  
+
+#### 5.2.1 **Option 1: Update notebook UI config before installing Kubeflow**
+Modify the file `awsconfigs/apps/jupyter-web-app/configs/spawner_ui_config.yaml`
```
goswamig, line 323: should be call out what modification is needed?

goswamig (commented):
Great work!!!

Minor comment.

docs/deployment/rds-s3/README.md
AlexandreBrown: The modification to be done is described before. If you asked this question it's probably because it's not clear enough.

I'll duplicate the change to make it clear.

AlexandreBrown: Done

AlexandreBrown:
I just did an end-to-end test and there is an issue, we cannot reference a secret that is not in the same namespace than the namespace we deploy the `PodDefault`.   
  
The secret `mlpipeline-minio-artifact` exists in the default user namespace (eg: `kubeflow-user-example-com`) therefore it works for aws access key and secret key values we can retrieve them as shown in the doc. But the secret `mysql-secret` is only in the `kubeflow` namespace therefore we cannot use it directly in our `PodDefault`.
  
If someone knows a better way to reference the secret `mysql-secret` let me know (maybe via the secret mounts?)

AlexandreBrown:
@goswamig @surajkota @akartsky @mbaijal 
I just did an end-to-end test.  
If we only add the S3 credentials to the pod default then it works as expected.  
If we also add the RDS credentials to the pod default then it doesn't work.  
The reason is, we cannot reference a secret that is not in the same namespace than the namespace we deploy the `PodDefault`.   
  
The secret `mlpipeline-minio-artifact` exists in the user namespace (eg: `kubeflow-user-example-com`) therefore it works for aws access key and secret key values we can retrieve as env variable in a notebook. But the secret `mysql-secret` is only in the `kubeflow` namespace therefore we cannot use it directly in our `PodDefault`.
  
If someone knows a better way to reference the secret `mysql-secret` let me know (maybe via the secret mounts?)

AlexandreBrown:
We also have to think about kfp sdk integration, currently users tend to use [`use_aws_secret`](https://kubeflow-pipelines.readthedocs.io/en/latest/_modules/kfp/aws.html) which relies on a secret.  
  
Is there a way to still use this with the secrets provider?

awsconfigs/apps/jupyter-web-app/configs/secrets-pod-default.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - name: database
+    valueFrom:
+      secretKeyRef:
+        name: mysql-secret
+        key: database
+  - name: port
+    valueFrom:
+      secretKeyRef:
+        name: mysql-secret
+        key: port
```
AlexandreBrown, line 47: After doing an end-to-end test I discovered that this part does not work since `mysql-sercret` is in `kubeflow` namespace (not user namespace eg: `kubeflow-user-example-com`) then we cannot use it as this `PodDefault` should be applied to the user's namespace (eg: `kubeflow-user-example-com`).  
  
It works for `mlpipeline-minio-artifact` secret since this secret is created in the user's namespace.  
  
Any thoughts/ideas/feedback would be greatly appreciated.

surajkota: I want to take a step back and ask why does user need access to database? If this was a common usecase I would expect the profile controller to take care of this


regarding the solution, I think you are in the right direction. the correct path would be to use the secrets provider to sync secrets from Secrets manager similar to how it is done in kubeflow namespace.

the secrets-manager component also makes this easy, user just need to change the namespace in kustomization.yaml and kustomize build and apply it

you are right accessing cross namespace wont work.

AlexandreBrown: You bring up a good point, the motivation is important.  
**Use Cases :**  
Here are the use cases that this PR aims to solve :  
- Be able to use `boto3` in a notebook and specify credentials without hard coding them into the notebook (access the credentials via env variables)
- Be able to use `ml-metadata` in a notebook (eg: to explore metadata, compare metrics) and specify the [credentials needed](https://github.com/google/ml-metadata/blob/master/g3doc/get_started.md) by using env variables
- Be able to use `ml-metadata` to query metadata during a pipeline run by using something similar to [`use_aws_secret`](https://kubeflow-pipelines.readthedocs.io/en/latest/_modules/kfp/aws.html) from kfp sdk. This is an important use case for me as we want to use `ml-metadata` to get back information about the current model in production, compare metrics to generate visualization during a pipeline run.
- Be able to use [`use_aws_secret`](https://kubeflow-pipelines.readthedocs.io/en/latest/_modules/kfp/aws.html) from kfp sdk to authenticate pipeline to push artifacts to s3 (artifacts are pushed automatically by kfp sdk v2).  
  
**About The Soltuion**  
>he correct path would be to use the secrets provider to sync secrets from Secrets manager similar to how it is done in kubeflow namespace.  
> the secrets-manager component also makes this easy, user just need to change the namespace in kustomization.yaml and kustomize build and apply it   

I'm not super familiar with `SecretProviderClass`, from my understanding you are suggesting we could create a new "empty" secret named `mysql-secret` in the user namespace, then we'd update the namespace for `secrets-mount.yaml`, `secrets-manager.yaml` and  `kustomization.yaml` to the user's namespace, then we could do kustomize build and apply and the secrets provider will be able to update the `mysql-secret` with the correct values in the user namespace?

AlexandreBrown: I have applied your suggestion @surajkota , see the updated changes.

AlexandreBrown:
@goswamig @surajkota

AlexandreBrown:
PR is ready for review, no longer WIP. @goswamig @surajkota

docs/deployment/rds-s3/README.md
```diff
@@ -286,7 +286,81 @@ mysql> select * from observation_logs;
...
+kubectl rollout restart deployment jupyter-web-app-deployment -n kubeflow
+```
+### 5.3 Verify notebook credentials configuration  
+To verify that your `PodDefault` setup was done successfully, navigate to the notebook creation UI and make sure you can select the `PodDefault`.  
+![](https://user-images.githubusercontent.com/26939775/155630906-0eecf1d9-3fb1-4d01-a85e-1cff46dc37e9.png)  
+Then create a notebook and check that the environment variables are accessible :  
+```python
+import os
+
+print(os.environ['AWS_ACCESS_KEY_ID'])
```
surajkota, line 359: nit: database port instead of sensitive information

docs/deployment/rds-s3/README.md
```diff
@@ -286,7 +286,81 @@ mysql> select * from observation_logs;
...
+Note that `YOUR_NAMESPACE` represents the namespace the secrets will be setup for.  
+For instance if your notebooks and pipelines will be in `kubeflow-user-example-com`, then you would use `kubeflow-user-example-com` in place of `YOUR_NAMESPACE`.
+```shell
+python secrets-setup.py --region YOUR_CLUSTER_REGION --cluster YOUR_CLUSTER_NAME --namespace YOUR_NAMESPACE
+```  
+You can always do `python secrets-setup.py --help` to learn more about the parameters available.    
+
+**Note :** The namespace must exist before executing the script.   
+
+### 5.2 (Optional) Set PodDefault selected by default in notebook UI
```
surajkota, line 319: Is there a generic existing Kubeflow documentation we can point to for this section?

docs/deployment/rds-s3/README.md
```diff
@@ -286,7 +286,81 @@ mysql> select * from observation_logs;
...
 
-## 5.0 Uninstall Kubeflow
+## 5.0 (Optional) Setup RDS & S3 credentials access for notebooks & pipelines
+### Use Cases
+- Be able to use `boto3` or aws libraries that require to pass in credentials in a notebook and specify credentials without hard coding them. Access the credentials via environment variables.  
+- Be able to use [ml-metadata](https://github.com/google/ml-metadata/blob/master/g3doc/get_started.md) in a notebook (eg: to explore metadata) and specify the credentials needed by using environment variables.
+- Be able to use [ml-metadata](https://github.com/google/ml-metadata/blob/master/g3doc/get_started.md) to query metadata during a pipeline run by passing a kubernetes secret to a pipeline component.  
+- Be able to use `boto3` or aws libraries that require to pass in credentials in a Kubeflow Pipeline Component. 
+
+The following script takes care of setting up secrets for notebooks and pipelines with values from the secrets manager.
```
surajkota, line 296: suggest some rewording to clarify things:

The following script creates a kubernetes `mysql-secret` and `mlpipeline-minio-artifact` secrets with RDS & S3 credentials specified in AWS Secrets Manager created while deploying the platform. This is a sample for demonstrating how you can use [PodDefault resource](https://github.com/kubeflow/kubeflow/blob/master/components/admission-webhook/README.md) and secrets in Notebooks to access the metadata database and and artifacts in S3 bucket created by pipelines. Make sure you create separate database and IAM users and corresponding secrets in Secrets Manager for your users if you want fine grain access control and auditing.

surajkota (changes requested):
Thanks for the investigation. Added a few comments for improving documentation and scripts

docs/deployment/rds-s3/secrets/secrets-setup.py
```diff
@@ -0,0 +1,136 @@
...
+    required=True
+)
+parser.add_argument(
+    '--namespace',
+    type=str,
+    metavar="USER_NAMESPACE",
+    help=f"The namespace you want to setup secrets for (eg: kubeflow-user-example-com)",
+    required=True
+)
+args, _ = parser.parse_known_args()
```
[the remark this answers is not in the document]
AlexandreBrown: Done

docs/deployment/rds-s3/secrets/secrets-setup.py
```diff
@@ -0,0 +1,136 @@
...
+        ])
+
+def setup_secrets_manager_kustomize():
+    print("=================================================================")
+    print("                     Secrets Manager Setup")
+    print("=================================================================")
+    secrets_manager_kustomize_content = read_secrets_manager_kustomize_content()
+    DEFAULT_NAMESPACE = secrets_manager_kustomize_content["namespace"]
+    edit_secrets_manager_kustomize(secrets_manager_kustomize_content, NAMESPACE)
+    build_secrets_manager_kustomize()
```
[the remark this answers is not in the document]
AlexandreBrown: Done

docs/deployment/rds-s3/secrets/secrets-setup.py
```diff
@@ -0,0 +1,136 @@
...
+)
+parser.add_argument(
+    '--cluster',
+    type=str,
+    metavar="CLUSTER_NAME",
+    help='Your cluster name (eg: mycluster-1)',
+    required=True
+)
+parser.add_argument(
+    '--namespace',
```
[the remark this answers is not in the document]
AlexandreBrown: Done

docs/deployment/rds-s3/secrets/secrets-setup.py
```diff
@@ -0,0 +1,136 @@
...
+    required=True
+)
+args, _ = parser.parse_known_args()
+if __name__ == "__main__":
+    CLUSTER_REGION = args.region
+    CLUSTER_NAME = args.cluster
+    NAMESPACE = args.namespace
+
+    AWS_CONFIGS_PATH = "../../../../awsconfigs/"
+    SECRETS_MANAGER_KUSTOMIZE_FILE_BASE_PATH = f"{AWS_CONFIGS_PATH}common/aws-secrets-manager/base/"
```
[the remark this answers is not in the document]
AlexandreBrown: Done

docs/deployment/rds-s3/secrets/secrets-setup.py
```diff
@@ -0,0 +1,136 @@
+import subprocess
```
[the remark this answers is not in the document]
AlexandreBrown: Thank you, I moved the script to `tests/e2e/utils/notebooks` as suggested and updated the doc.

docs/deployment/rds-s3/README.md
AlexandreBrown: You're right, done.

docs/deployment/rds-s3/README.md
AlexandreBrown: Thank you, done!

docs/deployment/rds-s3/README.md
```diff
@@ -286,7 +286,81 @@ mysql> select * from observation_logs;
 
 4. Verify the bucket is not empty and was populated by outputs of the experiment.
 
-## 5.0 Uninstall Kubeflow
+## 5.0 (Optional) Setup RDS & S3 credentials access for notebooks & pipelines
```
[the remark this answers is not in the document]
AlexandreBrown: Yes, done

goswamig (approved):

AlexandreBrown:
Ping @surajkota @goswamig

docs/component-guides/notebooks.md
```diff
@@ -0,0 +1,71 @@
...
+Note that `YOUR_NAMESPACE` represents the namespace the secrets will be setup for.  
+For instance if your notebooks and pipelines will be in `kubeflow-user-example-com`, then you would use `kubeflow-user-example-com` in place of `YOUR_NAMESPACE`.
+```shell
+PYTHONPATH=.. python utils/notebooks/setup_secrets_access.py --region YOUR_CLUSTER_REGION --cluster YOUR_CLUSTER_NAME --profile-namespace YOUR_NAMESPACE
+```  
+You can always do `PYTHONPATH=.. python utils/notebooks/setup_secrets_access.py --help` to learn more about the parameters available.    
+
+**Note :** The namespace must exist before executing the script.   
+
+### (Optional) Set PodDefault selected by default in notebook UI
```
surajkota, line 29: https://github.com/awslabs/kubeflow-manifests/pull/106#discussion_r822157387

Can you help create a PR to the main Kubeflow docs to address this? I agree the upstream docs only talk about how to make some configurations default

surajkota (changes requested):
Thanks for the PR!
Requesting a few cosmetic changes

docs/component-guides/notebooks.md
surajkota: I just realized that we should stick to having this configuration optional same as the Kubeflow docs as this change will introduce a dependency for this `PodDefault` to be available in all profile namespaces. Existing profile namespaces which do not have this resource will see failure in new notebook server creation and same for new namespaces

AlexandreBrown: @surajkota We can make a PR to update the kubeflow doc but it might be less intuitive for the user if we remove it from aws doc.  
The kubeflow doc will not be ableto give precise instructions on the value to add for aws since the pod default label would be different on GCP vs AWS for instance.  
Maybe we can add it to both, so have a more generic doc in Kubeflow doc, then to keep the current doc with the preicse label name so the user doesn't have to know too much about k8s or doesn't have to scan the pod default file to find the label that is applied in the script to get it setup.  
  
What do you think?

tests/e2e/utils/notebooks/common.py
```diff
@@ -0,0 +1,5 @@
+def print_banner(step_name: str):
```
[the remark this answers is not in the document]
AlexandreBrown: Done

tests/e2e/utils/notebooks/setup_secrets_access.py
```diff
@@ -0,0 +1,99 @@
...
+    setup_secrets_manager_service_account()
+    setup_secrets_manager_kustomize()
+    setup_notebook_secrets()
+
+    print_banner("AWS Secrets Setup Complete")
+
+def setup_secrets_manager_service_account():
+    print_banner("IAM Service Account Setup")
+
+    subprocess.run([
```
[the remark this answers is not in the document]
AlexandreBrown: Nice didn't know about this one, done!

tests/e2e/utils/notebooks/setup_secrets_access.py
```diff
@@ -0,0 +1,99 @@
...
+
+def edit_secrets_manager_kustomize(edit_secrets_manager_kustomize_content, namespace):
+    edit_secrets_manager_kustomize_content["namespace"] = namespace
+    write_yaml_file(edit_secrets_manager_kustomize_content, SECRETS_MANAGER_KUSTOMIZE_FILE_PATH)
+
+def setup_notebook_secrets():
+    print_banner("Kubeflow Notebook Setup")
+
+    print("Setting up PodDefault...")
+    subprocess.run([
```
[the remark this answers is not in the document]
AlexandreBrown: Good idea, done

docs/deployment/rds-s3/README.md
```diff
@@ -286,7 +286,7 @@ mysql> select * from observation_logs;
 
 4. Verify the bucket is not empty and was populated by outputs of the experiment.
 
-## 5.0 Uninstall Kubeflow
+## 6.0 Uninstall Kubeflow
```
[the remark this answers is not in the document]
AlexandreBrown: Good catch, done

docs/component-guides/notebooks.md
surajkota: Okay, for ease of use lets have this but add it to the doc:

Making this configuration default will introduce a dependency for the secrets and PodDefault to be available in all profile namespaces. Existing profile namespaces which do not have this resource will see failure in new notebook server creation if this PodDefault is not created

surajkota (changes requested):
requesting final change

awsconfigs/apps/jupyter-web-app/configs/pod-default.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: kubeflow.org/v1alpha1
+kind: PodDefault
```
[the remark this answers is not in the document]
AlexandreBrown: Done

docs/component-guides/notebooks.md
AlexandreBrown: Done

surajkota (approved):

PR #112, opened 2022-03-01
Fixed metadata db name hard coded to metadb

### LINKED ISSUE #110: Metadata db is not populated after running a pipeline
**Describe the bug**
When using Pipelines integration, if we specify a metadata db name that is different than the Kubeflow default of `metadb` then the metadata like artifacts types will still be pushed to a db named `metadb`.   

**Steps To Reproduce**
1. Specify a name different than `metadb` when setting up the rds db, let's use the db name `kubeflow` for this example
2. Install Kubeflow (tested using `main` branch + manifest `v1.4.1`)
3. Notice how the database `kubeflow` gets created correctly
image
image
4. Create a pipeline run that pushes some artifacts
Eg: 
```python
@component()
def test_comp(
    value: float,
    model: Output[Model],
    model_metrics: Output[Metrics]
):
    print(f"model_metrics.path {model_metrics.path}")
    print(f"model.path {model.path}")
        
    model_metrics.log_metric("test_loss", value)

    with open(model_metrics.path, 'w') as metrics_file:
        metrics_file.write(str(model_metrics.metadata))
    with open(model.path, 'w') as model_file:
        model_file.write("Some model data")
```
5. Verify that the artifacts metadata was pushed to `kubeflow` db
6. Notice how the db `kubeflow` does not have a new row in `Artifacts`
image
7. Verify the table `Artifacts` in `metadb` database instead
image

8. Notice how the metadata was pushed to the `metadb` database instead.  

**Expected behavior**
I expected the metadata to be uploaded to the `kubeflow` db.

**Environment**
* Kubernetes version 1.21
* Using EKS (yes/no), if so version? yes eks.4
* AWS service targeted (S3, RDS, etc.) RDS

**Screenshots**
image

**Additional context**
I could be wrong but I suspect that some changes are required to allow the renaming of the metadata db name.  
If we take a look at `upstream/apps/pipeline/upstream/base/metadata/base/metadata-grpc-deployment.yaml` we see the following :  
```yaml
env:
- name: DBCONFIG_USER
  valueFrom:
    secretKeyRef:
      name: mysql-secret
      key: username
- name: DBCONFIG_PASSWORD
  valueFrom:
    secretKeyRef:
      name: mysql-secret
      key: password
- name: MYSQL_DATABASE
  valueFrom:
    configMapKeyRef:
      name: pipeline-install-config
      key: mlmdDb
- name: MYSQL_HOST
  valueFrom:
    configMapKeyRef:
      name: pipeline-install-config
      key: dbHost
- name: MYSQL_PORT
  valueFrom:
    configMapKeyRef:
      name: pipeline-install-config
      key: dbPort
```  
So it looks like the `MYSQL_DATABASE` is retrieved from `upstream/apps/pipeline/upstream/base/installs/generic/pipeline-install-config.yaml`  
From this config map we can see that the value is hard coded  
```yaml
mlmdDb: metadb
```  
Maybe we'd need to add `mlmdDb` as a param in `awsconfigs/apps/pipeline/params.env` ?

AlexandreBrown:
Resolves #110  

Tested using custom db name `ml_metadata_db`   

Verified every databases and every tables content after executing a pipeline run with artifacts.  

**Checklist:**
- [x] Unit tests pass:
    2. `make test`
- [x] end-to-end test (new cluster, setup rds using automated script, create pipeline with artifacts, verify db and tables)

### DIFF
--- a/docs/deployment/rds-s3/auto-rds-s3-setup.py
+++ b/docs/deployment/rds-s3/auto-rds-s3-setup.py
@@ -564,6 +564,9 @@ def get_updated_pipeline_params_env_lines(db_instance_info, pipeline_params_env_
     db_host_pattern = "dbHost="
     db_host_new_line = db_host_pattern + db_instance_info["Endpoint"]["Address"] + "\n"
 
+    db_mlmd_db_pattern = "mlmdDb="
+    db_mlmd_db_new_line = db_mlmd_db_pattern + db_instance_info["DBName"] + "\n"
+
     bucket_name_pattern = "bucketName="
     bucket_name_new_line = bucket_name_pattern + S3_BUCKET_NAME + "\n"
 
@@ -574,6 +577,7 @@ def get_updated_pipeline_params_env_lines(db_instance_info, pipeline_params_env_
 
     for line in pipeline_params_env_lines:
         line = replace_line(line, db_host_pattern, db_host_new_line)
+        line = replace_line(line, db_mlmd_db_pattern, db_mlmd_db_new_line)
         line = replace_line(line, bucket_name_pattern, bucket_name_new_line)
         line = replace_line(line, minio_service_region_pattern, minio_service_region_new_line)
         new_pipeline_params_env_lines.append(line)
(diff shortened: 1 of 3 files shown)


docs/deployment/rds-s3/auto-rds-s3-setup.py
```diff
@@ -645,12 +649,12 @@ def footer():
     help=f"Unique identifier for the RDS database instance. Default is set to {DB_INSTANCE_NAME_DEFAULT}",
     required=False
 )
-DB_NAME_DEFAULT = "kubeflow"
+DB_NAME_DEFAULT = "ml_metadata_db"
```
akartsky, line 652: [nit] can we keep the default same as the vanilla : `metadb`

akartsky (approved):
LGTM 
Thanks !

docs/deployment/rds-s3/auto-rds-s3-setup.py
AlexandreBrown: Sure!

AlexandreBrown: Done

Issue #122, opened 2022-03-07
Should users be allowed to sign themselves up using Cognito hosted UI


surajkota:
**Is your feature request related to a problem? Please describe.**

**Current Scenario:**
The userpools created in deployment docs requires the AWS account/userpool admin to create users in cognito with their username, email address and a temporary password. Username is the only required parameter whereas email id is mandatory for kubeflow multi-user to work.

Sign up page only asks for username and password and successfully creates a user with `UNCONFIRMED` account status. This user cannot login since the account is created in `UNCONFIRMED` status and requires the userpool admin to confirm the account. However, the current sign up page does not ask for email address and so the users created via this page are not useful. The post-sign up page also lands in the following error:
![Screenshot 2022-03-07 at 2 05 48 PM](https://user-images.githubusercontent.com/22246703/157127077-5c3e235a-515a-4b78-855d-a6a52650a9fc.png)

Irrespective to the above error, we need to take a step back and ask, **do customers want to allow users to sign themselves up? or strictly control creation of user accounts?**

**Describe the solution you'd like**
Solution 1: Disable the sign up button from the Cognito hosted-UI and let admins handle user creations. This is secure since the domain is hosted over internet and customers strictly want to control who has access to the platform.
In addition to this, make email-id required attribute, call out that it is a required user attribute in docs and that it should have `email` as the key in user claims.

**Describe alternatives you've considered**
Solution 2: Keep the sign up page and make email address a mandatory parameter instead of username in the sign up page. We would still keep the account confirmation control for admins for security reasons.

These settings are for demo purposes only since cognito also allows many other customizations/features like federated login etc. and customers will configure the userpool according to their needs.

surajkota:
Current customers who want to disable sign up from hosted UI can select option `Only allow administrators to create users` under General Settings/Policies. See more on [aws docs](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-admin-create-user-policy.html)

PR #130, opened 2022-03-11
enhance cognito deployment automation scripts

surajkota:
- Enhance cognito automation script to include ingress and alb controller configuration
- Removed CNAME entry for `*.default.example.com` since it is not needed. From what I understand we only need `kubeflow.platform.example.com` and not even `*.platform.example.com` since all the apps are paths under the domain name. e.g. `platform.example.com/pipelines`
- Add badges to the README for customers to easily identify which Kubeflow version is under development. Open to other suggestions
  - Updated PR template to include license attribution
- Resolves: #122 
  - made email as a option to use as username instead of having email only or username only
  - Disable self signup - only admins can create users
  - Make email a required attribute and add corresponding documentation
- updated e2e test readme
- fixed logging issues in e2e tests using the `pytest.ini` config. It was not printing logging logs and also some of the print statements
- fixed issue with ingress and load balancer cleanup in the cognito e2e test. deleting ingress is clean way to remove the ingress managed load balancer. Using cognito_resource_cleanup was leaving around some resources
 
**Testing:**
- [x] e2e tests pass
- [x] Manually tested cognito automation README


tests/e2e/utils/cognito_bootstrap/cognito_resources_cleanup.py
```diff
@@ -62,6 +64,13 @@ def clean_root_domain(domain_name, hosted_zone_id, subdomain_hosted_zone):
     except Exception:
         pass
 
+def delete_policy(arn: str, region: str):
+    try:
+        policy = IAMPolicy(arn=arn, region=region)
+        policy.delete()
+    except Exception:
+        pass
```
ryansteakley, line 72: Is there some information we can provide here besides just passing?

ryansteakley:
Great work, just a few questions. Allowed only the admin to create accounts seems fine to me, seems like it also prevents the url from leaking and having someone sign up themselves thousands of times maliciously. However what is the reasoning for requiring the email field however?

tests/e2e/utils/cognito_bootstrap/cognito_resources_cleanup.py
surajkota: This code is only suppressing the error. Exception message is printed in the `delete()` method

ryansteakley (approved):

rrrkharse (approved):

ryansteakley (approved):

PR #133, opened 2022-03-15
[main-branch] Enhance EFS Tests: Dynamic Provisioning and validate using KFP

mbaijal:
This PR ports the changes from this closed PR #96 (1.3 branch) to the main branch

 - [x] Adds a test for EFS dynamic provisioning using the automated script, `auto-efs-setup.py`
 - [x] added a parameter to the auto-install script to specify a path to the `sc.yaml` file since the path requirements change when running from tests - this can be customized a bit more so that I can use it for static provisioning as well. 
 - [x]  Enhance the EFS static and dynamic provisioning tests to also test the volume usage using 2 KFP pipelines - one writes to the volume and other reads from it. 
 
**TODO:**
- [ ] Static Provisioning also needs an automated script (separate PR)
- [x] The KFP test is currently taking very long to run on my cluster - about 40minutes sometimes - this has increased with the number of runs on my cluster so possibly a resource limitation - I need to debug this.
- [ ] Filesystem cleanup in dynamic provisioning - issue is getting the created filesystem id from the script needs a little work

**Testing:** - 
Test Logs are uploaded to the gist here - https://gist.github.com/mbaijal/ff13a3f6cdd50adcf0aaf1ce0c8c8861

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`


tests/e2e/utils/constants.py
```diff
@@ -4,6 +4,7 @@
 
 DEFAULT_HOST = "http://localhost:8080/"
 DEFAULT_USER_NAMESPACE = "kubeflow-user-example-com"
+DEFAULT_NAMESPACE = "kube-system"
```
rrrkharse, line 7: Would it make sense to rename this to something like `DEFAULT_SYSTEM_NAMESPACE` to avoid confusion with the `default` namespace?

docs/deployment/add-ons/storage/efs/auto-efs-setup.py
```diff
@@ -540,50 +519,58 @@ def footer():
...
     help=f"Default is set to {EFS_THROUGHPUT_MODE_DEFAULT}",
-    required=False
+    required=False,
+)
+DEFAULT_DIRECTORY_PATH = ""
+parser.add_argument(
+    "--directory",
+    type=str,
+    default=DEFAULT_DIRECTORY_PATH,
+    help=f"Specify the path to the source files if different. Default is set to empty.",
```
rrrkharse, line 572: Does this directory contain EFS specific source files or what more is it intended to contain?

tests/e2e/tests/test_storage_efs.py
```diff
@@ -63,6 +105,135 @@ def test_pvc_with_volume(
...
+        assert "fs-" in fs_id
+
+        CLAIM_NAME = dynamic_provisioning["efs_claim_dyn"]
+        claim_list = subprocess.check_output("kubectl get pvc -A".split()).decode()
+        assert CLAIM_NAME in claim_list
+
+        claim_entry = subprocess.check_output(
+            f"kubectl get pvc -n {DEFAULT_USER_NAMESPACE} {CLAIM_NAME} -o=jsonpath='{{.status.phase}}'".split()
+        ).decode()
+        assert "Pending" in claim_entry
```
mbaijal, line 200: In the most recent run, this is being shown as Bound at this point. @mbaijal check

goswamig: Does bound happens immediately? If not please add a retry with timeout. In case of timeout the test should fail.

tests/e2e/tests/test_storage_efs.py
```diff
@@ -63,6 +105,135 @@ def test_pvc_with_volume(
...
+
+        # TODO: The following can be put into a method or split this into different tests
+        # TODO: The following section needs more assertions
```
goswamig, line 203: For e.g. ?

tests/e2e/utils/constants.py
mbaijal: done.

docs/deployment/add-ons/storage/efs/auto-efs-setup.py
mbaijal: Yes it contains only EFS specific files. This parameter should ideally not be required by users. We need this to be able to specify the path to the sc.yaml file from the tests.

tests/e2e/fixtures/storage_efs_dependencies.py
```diff
@@ -229,23 +247,119 @@ def on_create():
...
+    def on_create():
+        # Run the automated script to create the EFS Filesystem and the SC
+        efs_auto_script_absolute_filepath = os.path.join(
+            os.path.abspath(sys.path[0]), "../" + efs_auto_script_filepath
+        )
+        st = os.stat(efs_auto_script_filepath)
+        os.chmod(efs_auto_script_filepath, st.st_mode | stat.S_IEXEC)
+        subprocess.call(
+            [
+                "python",
```
surajkota, line 311: when we plugin these scripts into tests, I would highly recommend natively integrating them by moving it under `e2e/utils/efs` directory instead of calling python script using subprocess. Following are the benefits:

- validate that `requirements.txt` for the script is upto date since tests and script use same file
- reduce a lot of code duplication related to creating clients, kustomize, kubectl, service accounts, printing banners which increases the readability. I think you can cut down the script by 30%+
  - see difference between [scriptv1](https://github.com/awslabs/kubeflow-manifests/blob/e4a2aa55e0e40c1228f45698ca3e7dbb6129934d/docs/deployment/rds-s3/secrets/secrets-setup.py) vs [scriptv2](https://github.com/awslabs/kubeflow-manifests/blob/4170afed2da01b5c13b01107b0688a7cd7583c00/tests/e2e/utils/notebooks/setup_secrets_access.py)
- standardisation of running in terms of different deployment docs
- potentially better error handling for tests
- some of the constants defined in this method can go into the respective file

tests/e2e/fixtures/storage_efs_dependencies.py
```diff
@@ -229,23 +247,119 @@ def on_create():
...
+
+        efs_claim_dyn["efs_claim_dyn"] = claim_name
+        efs_claim_dyn["file_system_id"] = file_system_id
+
+    def on_delete():
+        kubectl_delete(efs_pvc_filepath)
+        kubectl_delete(efs_sc_filepath)
+
+        # Get FileSystem_ID
+        efs_claim_dyn = metadata.get("efs_claim_dyn") or efs_claim_dyn
```
surajkota, line 343: do you think a part of this code can become a `efs_cleanup` script? so the docs can reuse it in the cleanup steps

tests/e2e/resources/pipelines/pipeline_read_from_volume.py
```diff
@@ -0,0 +1,43 @@
...
+read_volume_op = create_component_from_func(
+    read_volume, base_image="python", packages_to_install=["boto3"]
+)
+
+
+@dsl.pipeline(
+    name="Reading from Volume KFP Component",
+    description="Read from Volume",
+)
+def read_from_volume_pipeline(mount_path=MOUNT_PATH, claim_name=CLAIM_NAME):
```
surajkota, line 40: nice idea!!

tests/e2e/utils/utils.py
```diff
@@ -85,6 +85,21 @@ def wait_for(callback, timeout=300, interval=30):
             time.sleep(interval)
 
 
+def wait_for_kfp_run_succeeded_from_run_id(kfp_client, run_id):
```
surajkota, line 88: +1 now, I feel we should start aws_utils, kfp_utils, kube_utils at some point

tests/e2e/tests/test_storage_efs.py
```diff
@@ -63,6 +105,134 @@ def test_pvc_with_volume(
...
+
+        fs_id = dynamic_provisioning["file_system_id"]
+        assert "fs-" in fs_id
+
+        CLAIM_NAME = dynamic_provisioning["efs_claim_dyn"]
+        claim_list = subprocess.check_output("kubectl get pvc -A".split()).decode()
+        assert CLAIM_NAME in claim_list
+
+        claim_entry = subprocess.check_output(
+            f"kubectl get pvc -n {DEFAULT_USER_NAMESPACE} {CLAIM_NAME} -o=jsonpath='{{.status.phase}}'".split()
```
surajkota, line 197: fyi: we are trying to use the k8s python library for most of these kubectl things https://github.com/awslabs/kubeflow-manifests/blob/main/tests/e2e/utils/custom_resources.py

tests/e2e/tests/test_storage_efs.py
```diff
@@ -63,6 +105,134 @@ def test_pvc_with_volume(
...
+        metadata,
+        setup_dynamic,
+        kfp_client,
+        account_id,
+        dynamic_provisioning,
+    ):
+        driver_list = subprocess.check_output("kubectl get csidriver".split()).decode()
+        assert "efs.csi.aws.com" in driver_list
+
+        pod_list = subprocess.check_output("kubectl get pods -A".split()).decode()
```
surajkota, line 181: - wdyt about splitting this into pvc test vs checking the installation is as expected?
- shouldnt the assert for some status instead of just being in the list?
- we are trying to use the k8s python library for most of these kubectl things https://github.com/awslabs/kubeflow-manifests/blob/main/tests/e2e/utils/custom_resources.py

tests/e2e/tests/test_storage_efs.py
```diff
@@ -24,27 +38,55 @@
...
 
-class TestEFS:
+class TestEFS_Static:
     @pytest.fixture(scope="class")
-    def setup(self, metadata, kustomize, static_provisioning):
+    def setup(self, metadata, kustomize, port_forward, static_provisioning):
+        # Disable caching in KFP
+        # By default KFP will cache previous pipeline runs and subsequent runs will skip cached steps
+        # This prevents artifacts from being uploaded to s3 for subsequent runs
+        patch_body = unmarshal_yaml(DISABLE_PIPELINE_CACHING_PATCH_FILE)
```
surajkota, line 73: mind moving to a common fixture/method instead of copying?

tests/e2e/fixtures/storage_efs_dependencies.py
```diff
@@ -1,14 +1,17 @@
 import time
+from unicodedata import name
```
surajkota, line 2: nit: unused?

surajkota (changes requested):
+1, few comments

tests/e2e/tests/test_storage_efs.py
mbaijal: There was a bug, it's fixed now. Yes binding happens immediately, no retry necessary.

tests/e2e/tests/test_storage_efs.py
mbaijal: The RDS test has a a few which verify the pipeline and experiment name etc are the same as the ones created. The difference is the pipeline compile command I am using is different so I'll have to first get the pipeline config details and then run the assertions which seems useless here - so I can actually just remove this TODO.

tests/e2e/fixtures/storage_efs_dependencies.py
mbaijal: I did consider this. However, the idea here is to "test the script" the way the user will use it and not to automate the resource creation for tests. This automation was already in place using fixtures which is a more pythonic way of testing. If I call only specific methods from the script, I am not testing it in entirety. In that case the script need not be a part of this EFS integration at all and we should consider running separate tests for these user facing scripts.

surajkota: Fair point. I think its easy to find a balance here, one way is to create a high level driver function which both the test and script call once they have all arguments, but I am open to calling via subprocess as well

For most of the other issues, we move the script to utils. it'll be easy to remove code duplication, avoid the problem with multiple requirements file and also wouldnt need a `--directory` which can be confusing w.r.t what files and directory structure are needed here

tests/e2e/fixtures/storage_efs_dependencies.py
mbaijal: yes, will do.

tests/e2e/utils/utils.py
mbaijal: yea that sounds good.

tests/e2e/tests/test_storage_efs.py
mbaijal: noted, let me fix that.

tests/e2e/tests/test_storage_efs.py
mbaijal: fair point, let me do that.

tests/e2e/fixtures/storage_efs_dependencies.py
mbaijal: Actually this needs a little more work, will create a separate PR for this.

tests/e2e/tests/test_storage_efs.py
mbaijal: Addressed the 2nd and 3rd points. 
For the first one, yes I considered this as well - as mentioned in the TODO too. This should ideally be done in a slightly larger refactor splitting the whole test into static and dynamic test files since a lot of resources are recreated across both. Created another issue for this too -  https://github.com/awslabs/kubeflow-manifests/issues/139

tests/e2e/fixtures/storage_efs_dependencies.py
mbaijal: I moved the file to the tests directory and updated the docs README.

mbaijal: IMHO this auto script is cleaner in the docs directory. I have moved it for now but let's discuss further tomorrow.

mbaijal: moved to utils as discussed in standup.

surajkota (approved):

rrrkharse (approved):

PR #136, opened 2022-03-17
Improved RDS S3 script to use shared security group for RDS db instan…

### LINKED ISSUE #137: Avoid  Ec2 VPC SecurityGroups Quota When Setting Up RDS
When trying to setup Kubeflow using `main` + `v1.4.1` upstream, I ran into an issue at the db creation step.  
`boto3` returned the following error : `An error occurred when calling the CreateDBInstance operation: The request violates your limit on number of Ec2 VPC SecurityGroups associated with an instance`   
After hours of back and forth with the support, no solution was found (we increased a lot of quotas related to EC2, VPCs etc but nothing worked).
  
After some more debugging I found the following quota : 
The default quota for the number of VPC Security Groups that can be attached to an RDS instance is 5 by default and this quota is not adjustable :    
securitygroupquota
I verified the number of VPC Security Group that was associated to the db instance during the db instance creation step and ours was 6, so just 1 over the limit.

To ensure all nodes have access to the database, instead of associating all the security groups to the database, we can associate just 1 security group called `ClusterSharedNodeSecurityGroup` that is meant to share communication with all nodes.  
This will ensure that no quota limit is reached regardless of your node count.

AlexandreBrown:
Resolves #137  

- [x] End-to-end Test on fresh cluster using `main` + `v1.4.1` and `cognito-rds-s3` setup
- [x] Verify that nodes have access to the RDS DB (KFP can push artifacts, db is populated etc)

### DIFF
--- a/docs/deployment/rds-s3/auto-rds-s3-setup.py
+++ b/docs/deployment/rds-s3/auto-rds-s3-setup.py
@@ -309,7 +309,7 @@ def create_db_instance(rds_client):
     print("Creating DB instance...")
 
     vpc_ids = get_cluster_vpc_ids()
-    vpc_security_group_ids = get_vpc_security_group_ids(vpc_ids)
+    vpc_security_group_id = get_vpc_security_group_id(vpc_ids)
 
     db_root_password = get_db_root_password_or_generate_one()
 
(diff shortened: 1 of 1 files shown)


docs/deployment/rds-s3/auto-rds-s3-setup.py
```diff
@@ -372,7 +372,7 @@ def get_vpc_id(vpc):
     return list(map(get_vpc_id, vpcs))
```
mbaijal, line 374: Could you add a comment here explaining the context

mbaijal (approved):
Could you add comment to the script explaining why we are picking only one SG. Otherwise looks ok.

docs/deployment/rds-s3/auto-rds-s3-setup.py
AlexandreBrown: Done!

rrrkharse (approved):

surajkota:
We need to validate these changes with tests. @AlexandreBrown could you help us move this script to `tests/e2e/utils` similar to #106. Feel free to change the test code as well to use these script

we are working on PR builds which should help with automated tests

Issue #129, opened 2022-03-11
Deployment documentation is broken


tejarora:
**What is the URL of the document?**
Issues pertain to 3 documents (3 different branches)
1. https://github.com/awslabs/kubeflow-manifests/tree/main/docs/deployment/vanilla
2. https://github.com/awslabs/kubeflow-manifests/tree/release-v1.3.1-aws-b1.0.0/distributions/aws/examples/vanilla
3. https://github.com/awslabs/kubeflow-manifests/tree/v1.3-branch/distributions/aws/examples/vanilla

**Which section(s) is the issue in?**
See below

**What needs fixing and describe the solution you'd like?**
(a)
All 3 URLs above have a vague and incomplete step
"AWS IAM permissions to create roles and attach policies to roles."

(b) 
The *v1.3* branches have no configuration for kubeflow version, but the main branch does. In fact the main branch talks about cloning the kubeflow repo, but the branch documents do not. 
How does one one set kubeflow version using v1.3* branches? Which branch is prescribed for deployment?
This is a mess

(c) The document for the branch release-v1.3.1-aws-b* has a step that is asking to use a different branch v1.3-branch

(d) The documents for the *v1.3* branches have their deployment step as:
while ! kustomize build example
Is this deployment not for serious use? Is it just for experimentation and not ready for production??

(e) The *v1.3* documents talk about changing a password, but there is no file whatsoever on the branches that contains the password. There is reference to a dex/base/config-map.yaml but this file does not exist in the branches. It actually exists in the kubeflow repo, which the branch documentation does not talk about at all....

**Additional context**
I would like to use kubeflow on a production EKS cluster, but the only documentation I found to help with that is well short of expectations.

goswamig:
@tejarora  thanks for reporting this.

a) This is just a general pre-requisite that user/role which are trying to deploy the kubeflow has permission to create role and attach polices.  This is the IAM user/role who is deploying the kubeflow, which could be "you (mac)" if "you(mac)" are trying to deploy kubeflow on your eks cluster or it could be "your(Ec2)" instance from where you are deploying kubeflow.


b/c) the git clone instruction has checkout branch 
```
 git checkout v1.3-branch 
```

d) This step has been borrowed from official guide, however its basically retry the deployment if it fails. You can always opt for multiple steps deployment. 

[...]

PR #142, opened 2022-03-18
add releases and versioning documentation

### LINKED ISSUE #76: Create a master branch for AWS Distro Development
## Background

Current development process being followed for AWS-Kubeflow-distro-1.3.1 was to take a release branch from the manifest repository as the base (a.k.a upstream) and make AWS distribution specific changes on the same branch. This approach presented the fastest way to get development going considering the timeline, kfctl tool getting deprecated and vendors requiring to setup their own repository to host distribution specific manifests. 

## Problem

However, the above approach presents several issues:

1. Working on a new release:
    1. Every release requires a developer to create a copy of upstream KF release branch and copying/cherry-picking changes from the awslabs previous release branch (e.g. v1.3-branch) to new branch (e.g. v1.4-branch) which is cumbersome and error prone (e.g. #75)
    2. These release branches on upstream are not frozen because they are used for any patch releases. Let’s take an example for v1.4.0 and v1.4.1. Every patch release will require us to either:
        1. Create a new branch(e.g. v1.4.1-branch) from the tag/v1.4.1 from upstream and repeat the process in 1.a.
        2. Cherry pick commits related to patch release from upstream and create a PR to awslabs repo
        3. Rebase the awslabs/v1.4-branch from upstream/v1.4-branch and force push which is unacceptable
2. Maintenance issues unrelated to the code owned by the team
    1. Security vulnerabilities for test dependencies coming from upstream
    2. Modifying the root pkg READMEs on each release branch (e.g. [see README changes](https://github.com/awslabs/kubeflow-manifests/pull/42/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5))
3. AWS release version is coupled with Kubeflow version
    1. With this approach, we need to maintain 2 release versions one from Kubeflow and another from aws because we may have to do patch releases or feature enhancements e.g. aws-manifests-1.x-Kubeflow-1.3.1
4. Other issues:
    1. PR build and test infra(when we have one) will need to be constantly be updated to point to different branches instead of one master branch. Similarly Github issue templates need to be ported
    2. Default branch on the repository will keep changing
    3. Might loose the history related to the PRs between releases if changes are copied instead of cherry-picking

## Suggested Approach

Majority of the overlays do not change unless there are major changes in the upstream component. Create a base with AWS overlays that does not change between versions and work only on the delta per release.

### Development - Build and Test

Create a master branch to host only AWS specific overlays, tests and docs. e.g.

```
ubuntu@ip-172-31-0-119:~/kubeflow-manifests$ tree -L 3
.
├── README.md
├── awsconfig
│   ├── apps
│   │   ├── jupyter
│   │   ├── katib
│   │   └── pipelines
│   ├── common
│   │   ├── alb-controller
│   │   ├── istio-envoy-filter
│   │   └── istio-ingress
│   └── infra_configs
├── docs
│   ├── cognito
│   ├── rds
│   ├── s3
│   └── storage
├── tests
│   ├── canary
│   ├── e2e
│   └── unit

```

(**Customer experience)** This will introduce one additional step for users after cloning the awslabs repo.

```
export KUBEFLOW_RELEASE_VERSION=v1.4.1
git clone --branch ${KUBEFLOW_RELEASE_VERSION} https://github.com/kubeflow/manifests.git upstream
```

Once user has cloned a tagged branch from upstream manifests(kubeflow/manifests repo), they will continue with setting up aws resources or running kustomize build and apply just like today.

For this, we need to isolate the aws specific changes in form of overlays to a separate directory.

* I have already completed a [POC](https://github.com/awslabs/kubeflow-manifests/pulls?q=is%3Apr+author%3Asurajkota+v1.4) for jupyter, cognito, rds and s3 in the v1.4-branch as a POC. There should be no changes for EFS and FSx since the changes are self contained. The only piece left is secrets-manager integration which is also doable

### Releases

New releases (major and patch from upstream): Cut a release once a Kubeflow release has been validated with updated README. 

Patch for an old release(depends on support policy) where latest overlays are incompatible: Use the existing release tag branch to do a patch release or backport features for older releases.

* Follow up: Will this base become outdated between several releases?
    * The intent of this development strategy is not to support multiple versions using the latest build. Making changes to a previous release will take the same amount of time i.e. go to release branch and make the changes.

### How does this solve the problems stated earlier?

We take out the problems related to PR build/test, copying changes over from old release branches and modifying parent READMEs etc by creating this base package with only AWS manifests instead of using a Kubeflow released branch as base. The contributors only needs to work on the delta per release. Problems related to security vulnerabilities because of unrelated packages are handled by only cloning the upstream during build time.

Patch releases just become a small PR with changing a version in a config file and re-running the tests using the PR build infra.

The problem for release version is not completely solved since a it also depends on the upstream but we are able to cut a new release for this package independently which talks about the compatible version of Kubeflow (TBD - see below)

### Next steps/TODOs

* How to document/convery the supported upstream version to users?
* What is the release versioning for this package?
* Use the [main branch](https://github.com/awslabs/kubeflow-manifests/tree/main) to get started with this

Please comment your thoughts/feedback

surajkota:
Resolves #76, few questions from #129

- Add releases and versioning documentation and clarification in prerequisites
- Removes IAM instruction which does not add value. If we do need it, it needs to be many more policies and specific. Out of scope for this PR
- added dex configmap path relative to root of repo like other instructions

**Testing:**
- git clone works


docs/releases.md
```diff
@@ -0,0 +1,8 @@
...
+
+Since each AWS distribution release is built on top of a Kubeflow release, the releases follow the following format: `{KUBEFLOW_RELEASE_VERSION}-aws-b{BUILD_NUMBER}` where `KUBEFLOW_RELEASE_VERSION` refers to the [Kubeflow's released version](https://github.com/kubeflow/manifests/releases) and `BUILD_NUMBER` refers to AWS build for that Kubeflow version. `BUILD_NUMBER` uses [semantic versioning](https://semver.org/) (SemVer) to indicate whether changes included in a particular release introduce features or bug fixes and whether or not features break backwards compatibility.
+
+When a Kubeflow on AWS version is released, a Git tag containing a `{KUBEFLOW_RELEASE_VERSION}-aws-b{X.Y.Z}` version is created on the repository. These releases can be found under the repository's [releases](https://github.com/awslabs/kubeflow-manifests/releases) section.
```
ryansteakley, line 8: Any reason we refer to it as `b{X.Y.Z} `but previously we had `b{BUILD_NUMBER}`
This is in reference to the format of `{KUBEFLOW_RELEASE_VERSION}-aws-b{X.Y.Z}` vs `{KUBEFLOW_RELEASE_VERSION}-aws-b{BUILD_NUMBER}` which I believe refer to the same thing?

surajkota: good catch

ryansteakley (approved):

docs/deployment/cognito/README.md
```diff
@@ -32,16 +32,14 @@ This guide assumes that you have:
         --nodes-max 10 \
         --managed
         ```
-1. AWS IAM permissions to create roles and attach policies to roles.
 
-1. Clone the `awslabs/kubeflow-manifest` repo, `kubeflow/manifests` repo and checkout the desired branches
-    1. Substitute the value for `KUBEFLOW_RELEASE_VERSION`(e.g. v1.4.1) and `AWS_MANIFESTS_BUILD`(e.g. v1.4.1-b1.0.0) with the branch or tag you want to use below
+1. Clone the `awslabs/kubeflow-manifest` repo, `kubeflow/manifests` repo and checkout the release branches.
+    1. Substitute the value for `KUBEFLOW_RELEASE_VERSION`(e.g. v1.4.1) and `AWS_RELEASE_VERSION`(e.g. v1.4.1-aws-b1.0.0) with the tag or branch you want to use below. Read more about [releases and versioning](../../community/releases.md#releases-and-versioning) policy if you are unsure about what these values should be.
```
akartsky, line 37: [nit] can we change [line 37] numbered list to bullet list just like in [line 19 and 20] to be consistent

akartsky (approved):

PR #144, opened 2022-03-18
Added KFServing/KServe Service Account Support For Secrets Manager

### LINKED ISSUE #143: Make secrets created by Secrets Manager work with KServe/KFServing Service Account
KFServing/KServe allows you to use a Service Account that references a secret in order to pull a model stored on S3 for inference.  
By default KFServing/KServe expects the secrets keys to be `AWS_ACCESS_KEY_ID` and `AWS_SECRET_KEY` but the keys created from the secrets created by the Secrets Manager integration are `accesskey` and `secretkey`.  
This results in KFServing/KServe (the storage initializer) being unable to pull the model from S3.  

We should use Kustomize and add an overlay to edit the `inferenceservice` definition to change the key names.  
I will create a separate PR on KServe doc to update it for aws.

AlexandreBrown:
Resolves #143 

**Testing:**
- [x] Unit tests pass
- [x] e2e tests pass

### DIFF
--- a/awsconfigs/apps/kserve/inferenceservice.yaml
+++ b/awsconfigs/apps/kserve/inferenceservice.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: inferenceservice-config
+  namespace: kubeflow
+data:
+  credentials: |-
+    {
+       "s3": {
+           "s3AccessKeyIDName": "accesskey",
+           "s3SecretAccessKeyName": "secretkey"
+       }
+    }
--- a/docs/component-guides/kserve.md
+++ b/docs/component-guides/kserve.md
@@ -1,0 +1,70 @@
+# Use S3 as storage Uri for KFServing/KServe
+  
+## Motivation  
+When deploying a model server, you can specify in the `storageUri` the S3 path to your model.  
+If the S3 bucket is private, KFServing/KServe will need to be authorized to download your model.  
+To do so, we can leverage the fact that KFServing/KServe supports Kubernetes [Service Accounts](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#:~:text=A%20service%20account%20provides%20an,recommended%20by%20the%20Kubernetes%20project.).  
+
+## Overview  
+The goal is to create a Service Account that has access to the bucket and then specify this Service Account in the inference service definition to allow KFServing/KServe to download the model.    
+The Service Account will reference the secret with AWS credentials created during the installation of Kubeflow, more precisely, it will reference the secret `mlpipeline-minio-artifact` created by the Secrets Manager.  
+  
+## 1.0 Prerequisites  
+  
+1. The secrets were setup for the namespace you want to deploy your model. This can be done by following [this guide](./notebooks.md#setup-the-secrets-access)
+  
+## 2.0 Setup  
+1. Create a Service Account definition that references the secret `mlpipeline-minio-artifact`  .  
+
+**aws-service-account.yaml**  
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: aws-sa
+secrets:
+- name: mlpipeline-minio-artifact  
+```  
+
+2. Apply the Service Account to the namespace you want to deploy your model to (eg: kubeflow-user-example-com).    
+```
+kubectl apply -f aws-service-account.yaml -n <YOUR_NAMESPACE>  
+```  
+For instance, if your namespace is `kubeflow-user-example-com`, then you would do the following :  
+```
+kubectl apply -f aws-service-account.yaml -n kubeflow-user-example-com  
+```  
+  
+3. Specify the Service Account when deploying a model server  
+This can be done by adding it to the `InferenceService` yaml definition :  
+**sklearn_iris.yaml**
+```yaml
+apiVersion: "serving.kubeflow.org/v1beta1"
+kind: "InferenceService"
+metadata:
+  name: "sklearn-iris"
+spec:
+  predictor:
+    serviceAccountName: aws-sa
+    sklearn:
+      storageUri: <YOUR_S3_URI>  
+```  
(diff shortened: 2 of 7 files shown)


awsconfigs/apps/kserve/inferenceservice.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: inferenceservice-config
+  namespace: kubeflow
+data:
+  credentials: |-
+    {
+       "s3": {
```
goswamig, line 9: Have you tried using [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). I believe if you're using in production its better to use IAM role than static credential. 

There has been some discussion around [here](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) however I am not sure if the feature works in KServe. It would be worth testing it.

goswamig (commented):
Though changes looks good to me. I would suggest to verify the IRSA setup instead of static credentials.

awsconfigs/apps/kserve/inferenceservice.yaml
AlexandreBrown: @goswamig I have not tried using IRSA with KFServing/KServe, I know KFServing/KServe expects the AWS environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`) to be available for the `storage-initializer`, not sure if IRSA works or how to set it up. I agree that long term IRSA is the best approach, but it doesn't seem to be fully documented yet.  

I went for the minimal approach here since currently [KServe documentation](https://kserve.github.io/website/modelserving/storage/s3/s3/) only showcase one way (via secret + service account creation)

surajkota:
@AlexandreBrown I want to propose instead of changing the secret key names in kserve deployment, we should aim to change the keys in `mlpipeline-minio-artifact` for 2 reasons:
- `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are the default names used for credentials in [AWS documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html). Current variable names in `mlpipeline-minio-artifact` deviate from the convention that is normally followed.
- The Key names in `inferenceservice` configmap are user facing and it is already in use from [kserve 0.4.0 onwards](https://github.com/kserve/kserve/issues/999#issuecomment-678884150). Changing this will require existing users to  change their configmaps.

 @rrrkharse, @AlexandreBrown do you think this is doable?

docs/component-guides/kserve.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Overview  
+The goal is to create a Service Account that has access to the bucket and then specify this Service Account in the inference service definition to allow KFServing/KServe to download the model.    
+The Service Account will reference the secret with AWS credentials created during the installation of Kubeflow, more precisely, it will reference the secret `mlpipeline-minio-artifact` created by the Secrets Manager.  
+  
+## 1.0 Prerequisites  
+  
+1. The secrets were setup for the namespace you want to deploy your model. This can be done by following [this guide](./notebooks.md#setup-the-secrets-access)
+  
+## 2.0 Setup
```
surajkota, line 16: everything is very well described in kserve docs here: https://kserve.github.io/website/modelserving/storage/s3/s3/. What value are we adding by having these instructions here? Trying to avoid duplication because there is a high chance one of them can eventually get outdated.

I think we should give a brief summary and redirect to Kserve documentation

AlexandreBrown: @surajkota I was not sure if we should put this in AWS repo or in KServe doc.  
It seems like dsun suggested we update KServe doc (for context see https://kubeflow.slack.com/archives/CH6E58LNP/p1647567321171729?thread_ts=1647566004.793539&cid=CH6E58LNP)  
  
So I think you're right, we should keep it simple here and redirect to KServe doc and update KServe doc with whatever approach is taken.

AlexandreBrown:
> @AlexandreBrown I want to propose instead of changing the secret key names in kserve deployment, we should aim to change the keys in `mlpipeline-minio-artifact` for 2 reasons:
> 
> * `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are the default names used for credentials in [AWS documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html). Current key names in `mlpipeline-minio-artifact` deviate from the convention that is normally followed.
> * The Key names in `inferenceservice` configmap are user facing and it is already in use from [kserve 0.4.0 onwards](https://github.com/kserve/kserve/issues/999#issuecomment-678884150). Changing this will require existing users to  change their configmaps.
> 
> @rrrkharse, @AlexandreBrown do you think this is doable?

I totally agree with your proposal.

awsconfigs/apps/kserve/inferenceservice.yaml
goswamig: Right, what I am saying is instead of creating secrets and passing its name to service account. Can we rather just annotate the service account with a role which has policy attached for S3 download/upload? This will help us to at least rule out the feature is broken.

AlexandreBrown: @goswamig I'm open to it for sure, I'd just need more guidance.  
  
Would it be [something like this ](https://aws.amazon.com/premiumsupport/knowledge-center/eks-restrict-s3-bucket/?nc1=h_ls)?

goswamig: Given that user (under profiler) will create `InferenceService`, you can annotate your profile with IAM service which will have attached policy to READ/Write S3. 

For e.g. https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/config/samples/profile_v1_aws_iam.yaml#L9

AlexandreBrown:
Closing this issue as it has been stale for a long time and won't have time to work on it soon.  
  
Feel free to re-open and work on it @surajkota @goswamig to add IRSA

PR #152, opened 2022-03-28
Revert "Decouple RDS and S3 KFP installation"

rrrkharse:
Reverts awslabs/kubeflow-manifests#108

The commit couples the secret manager integration with the KFP installation, which is an issue because Katib (a separate component) is also dependent on secret manager.


ryansteakley (approved):

surajkota (approved):

goswamig:
i would recommend to work on fix than reverting it.  Unless fix is impossible or will take lots of time.

rrrkharse:
While the changes to be made are not major review of the fix could take a day or two depending on feedback. I'll revert the original PR and target a follow up PR asap.

PR #153, opened 2022-03-28
Add a script to automate FSX resource creation

mbaijal:
 - [x]  Add a script to automate FSX resource creation
 
**Testing:** - 

Link to testing logs - https://gist.github.com/mbaijal/1fec71cd7f4d421ccf007f074b110c28

**Checklist:**
- [ ] Unit tests pass:
    2. `make test`

### DIFF
--- a/tests/e2e/fixtures/storage_fsx_dependencies.py
+++ b/tests/e2e/fixtures/storage_fsx_dependencies.py
@@ -2,6 +2,7 @@ import time
 import pytest
 import subprocess
 import boto3
+import os, stat, sys
 
 from e2e.utils.config import metadata
 from e2e.utils.cognito_bootstrap.common import load_cfg, write_cfg
@@ -31,16 +32,6 @@ from e2e.utils.constants import (
 )
 
 
-def wait_on_fsx_status(desired_status, fsx_client, file_system_id):
-    def callback():
-        response = fsx_client.describe_file_systems(FileSystemIds=[file_system_id])
-        filesystem_status = response["FileSystems"][0]["Lifecycle"]
-        print(f"{file_system_id} {filesystem_status} .... waiting")
-        assert filesystem_status == desired_status
-
-    wait_for(callback, 600)
-
-
 def get_fsx_dns_name(fsx_client, file_system_id):
     response = fsx_client.describe_file_systems(FileSystemIds=[file_system_id])
     return response["FileSystems"][0]["DNSName"]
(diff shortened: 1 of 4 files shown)


tests/e2e/fixtures/storage_fsx_dependencies.py
```diff
@@ -51,179 +42,63 @@ def get_fsx_mount_name(fsx_client, file_system_id):
...
+        return filesystem["Name"] == file_system_name
 
-    def on_create():
-        kubectl_apply_kustomize(FSx_CSI_DRIVER)
-        fsx_driver["driver_version"] = FSx_DRIVER_VERSION
+    file_systems = fsx_client.describe_file_systems()["FileSystems"]
 
-    def on_delete():
-        kubectl_delete_kustomize(FSx_CSI_DRIVER)
+    file_system = next(filter(name_matches, file_systems))
```
ryansteakley, line 51: Can you explain this line to me? I see that we are filtering for a name_match which will return an iterable but why are we calling next here?

ryansteakley, line 51: Can you explain this line to me? I see that we are filtering for a name_match which will return an iterable but why are we calling next here?

tests/e2e/fixtures/storage_fsx_dependencies.py
```diff
@@ -51,179 +42,63 @@ def get_fsx_mount_name(fsx_client, file_system_id):
...
-    def on_delete():
-        # Get FileSystem_ID
-        details_fsx_volume = metadata.get("fsx_volume") or fsx_volume
-        fs_id = details_fsx_volume["file_system_id"]
-        sg_id = details_fsx_volume["security_group_id"]
-
-        # Delete the Filesystem, does not provide a deleted status hence can't check delete status
-        fsx_client.delete_file_system(
-            FileSystemId=fs_id,
+        st = os.stat(fsx_auto_script_filepath)
```
ryansteakley, line 74: does st refer to stat?

ryansteakley, line 74: does st refer to stat?

tests/e2e/fixtures/storage_fsx_dependencies.py
```diff
@@ -51,179 +42,63 @@ def get_fsx_mount_name(fsx_client, file_system_id):
...
+                "--write_to_file",
+                "True",
+            ]
         )
 
-        # Delete the Security Group
-        ec2_client.delete_security_group(GroupId=sg_id)
+        with open("fsx-config.txt", "r") as f:
+            file_system_id = f.read()
+        # file_system_id = get_file_system_id_from_name(fsx_client, claim_name)
```
ryansteakley, line 95: nit: left over comment?

ryansteakley, line 95: nit: left over comment?

tests/e2e/fixtures/storage_fsx_dependencies.py
```diff
@@ -236,16 +111,25 @@ def on_create():
...
 
         fsx_claim["claim_name"] = claim_name
+        fsx_claim["file_system_id"] = file_system_id
 
     def on_delete():
         kubectl_delete(fsx_pvc_filepath)
         kubectl_delete(fsx_pv_filepath)
 
+        details_fsx_volume = metadata.get("fsx_claim") or fsx_claim
+        # file_system_id = details_fsx_volume["file_system_id"]
```
ryansteakley, line 125: Left over comment block?

ryansteakley, line 125: Left over comment block?

tests/e2e/utils/auto-fsx-setup.py
```diff
@@ -0,0 +1,478 @@
...
+    footer()
+
+
+def write_to_file(fs_id):
+    if WRITE_TO_FILE or WRITE_TO_FILE == "True":
+        with open("fsx-config.txt", "w") as fh:
+            fh.write(fs_id)
+
+
+def rand_name(prefix):
```
ryansteakley, line 31: Can you use https://github.com/awslabs/kubeflow-manifests/blob/6d72c6c115a9e9b7bec89953586a3c9b982b3acc/tests/e2e/utils/utils.py#L102 ?

tests/e2e/fixtures/storage_fsx_dependencies.py
mbaijal: yes, cleaned up and also cleaned the extra method that is now not being used.

tests/e2e/fixtures/storage_fsx_dependencies.py
mbaijal: no actually, has commented for some testing. Uncommented it.

tests/e2e/utils/auto-fsx-setup.py
mbaijal: Technically yes, but for now we have kept the auto script standalone thus this one method is copied over. This should change very soon in the next iteration of the scripts.

tests/e2e/fixtures/storage_fsx_dependencies.py
mbaijal: that was some bad variable naming, fixed it.

tests/e2e/fixtures/storage_fsx_dependencies.py
mbaijal: This code was not being used at all, as you pointed out below. Removed the code block.

ryansteakley (approved):

ryansteakley:
Looks good, script is working with revisions made ?

mbaijal:
> 

Running ...

mbaijal:
> Looks good, script is working with revisions made ?

Test is running, will update.

tests/e2e/utils/auto-fsx-setup.py
goswamig: I would recommend to reuse the code than copying entire method.

tests/e2e/utils/auto-fsx-setup.py
```diff
@@ -0,0 +1,478 @@
...
+            tag["Key"] == "alpha.eksctl.io/cluster-name"
+            and tag["Value"] == CLUSTER_NAME
+            for tag in oidc_provider_tags
+        ):
+            return True
+
+    return False
+
+
+def get_iam_client():
```
goswamig, line 91: please remove all the copied code and re-use from existing file.

tests/e2e/utils/auto-fsx-setup.py
```diff
@@ -0,0 +1,478 @@
...
+        status = fsx_client.describe_file_systems(FileSystemIds=[file_system_id])[
+            "FileSystems"
+        ][0]["Lifecycle"]
+
+        if status == "error":
+            raise Exception(
+                "An unexpected error occurred while waiting for the fsx file system to become available!"
+            )
+
+        sleep(10)
```
goswamig, line 384: why 10 seconds?

goswamig (changes requested):
I would suggest to not to duplicate the code but rather re-use from existing file.

tests/e2e/utils/auto-fsx-setup.py
mbaijal: I agree with your point and this has already been discussed in the team. However - 
1. This change is needed across all auto-scripts and is a larger refactor which is already in the plans. 
2. Making any changes to the current scripts seems like a trivial task but ends up breaking things and takes longer than expected and IMO is beyond the scope of this PR. 
3. And most importantly **the goal of these scripts is that they are meant for the user to use as standalone scripts** which will not be the case if I start re-using methods from utils. If this is not a current goal, this PR is not needed at all and the existing fixtures serve our purpose well enough. 

Based on the above, we can either - 
1. Merge this PR as is, and then follow up script refactoring across all automated scripts in a different story
2. or, close this PR until we have the bandwidth to rewrite the scripts, which is also ok since Suraj had a similar point. 

Let me know your thoughts @goswamig.

mbaijal:
> > Looks good, script is working with revisions made ?
> 
> Test is running, will update.

Test completed successfully.

tests/e2e/utils/auto-fsx-setup.py
```diff
@@ -0,0 +1,478 @@
...
+def does_need_to_create_fsx_iam_policy():
+    iam_resource = get_iam_resource()
+    try:
+        iam_resource.Policy(FSX_IAM_POLICY_ARN).load()
+        return False
+    except iam_resource.meta.client.exceptions.NoSuchEntityException:
+        return True
+
+
+def get_iam_resource():
```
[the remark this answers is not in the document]
mbaijal: Let's discuss above https://github.com/awslabs/kubeflow-manifests/pull/153#discussion_r839988238

tests/e2e/utils/auto-fsx-setup.py
mbaijal: I have removed most common code which was being duplicated. However, the script can be cleaned even further such as some of the utilities that are only needed for FSx filesystem creation like subnet etc are still in this file. We are planning a larger refactor for some of this code as discussed last week and a lot of this code is expected to change in the next iteration.

tests/e2e/utils/auto-fsx-setup.py
mbaijal: I have removed most common code which was being duplicated. However, the script can be cleaned even further such as some of the utilities that are only needed for FSx filesystem creation like subnet etc are still in this file. We are planning a larger refactor for some of this code as discussed last week and a lot of this code is expected to change in the next iteration.

tests/e2e/fixtures/storage_fsx_dependencies.py
```diff
@@ -236,16 +96,25 @@ def on_create():
...
         kubectl_delete(fsx_pv_filepath)
 
+        details_fsx_volume = metadata.get("fsx_claim") or fsx_claim
+        file_system_id = details_fsx_volume["file_system_id"]
+
+        fsx_client.delete_file_system(
+            FileSystemId=file_system_id,
+        )
+
+        # TODO: Find a way to delete the security group
```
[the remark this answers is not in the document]
mbaijal: This was a miss on my part since earlier it was difficult to share variables across scripts, this is being done via config files now. Added the cleanup.

tests/e2e/utils/auto-fsx-setup.py
```diff
@@ -0,0 +1,458 @@
...
+    if is_prerequisite_met:
+        print("eksctl found!")
+    else:
+        raise Exception(
+            "Prerequisite not met : eksctl could not be found, make sure it is installed or in your PATH!"
+        )
+
+
+def verify_kubectl_is_installed():
+    print("Verifying kubectl is installed...")
```
goswamig, line 96: I would encourage to put such generic function in utils file. 
also code for such functions can be combined in just one function called `verify_prerequisite_is_installed()` and you can check for all such pre-requisite once.

tests/e2e/utils/auto-fsx-setup.py
goswamig: the concern is random amount of sleep.

goswamig (approved):

PR #156, opened 2022-03-30
Unit-tests for each deployment option

ryansteakley:
Adds unit-tests for each AWS Kubeflow-manifests deployment option.
- Added in `generate_tests.py`
Additional PR on the way which adds enhancement for substitutions of fields such as those supplied to params.env

Made modifications to `test_util.go` to run unit-tests as caching system relied on just metadata name and kind but unit-tests would fail since k-eventing and k-serving both have the same metadata name and kind.

- Added namespace to the key that is constructed as well

**Testing:**
- [x] Unit tests pass
Screen Shot 2022-03-30 at 7 23 37 PM

- [ ] e2e tests pass


ryansteakley:
> Intimidating PR, can you link where this unit test is coming from and result of test run/log?

Hey these unit-tests are generated by `test_util.go` so they have not existed prior.

mbaijal (approved):

surajkota:
Discussed having parametrized tests for components which have params.env and only check the kustomize file builds for each deployment option since the validation is not maintainable by this group. This is because many components are being consumed as is from upstream. We will focus on validating AWS managed overlays and components

surajkota:
Reopen PR when ready

PR #163, opened 2022-04-07
Bug fix in fsx-auto script

mbaijal:
 - Found a bug in the FSx script when I ran it on a new cluster.
 - Also, made the policy name variable randomized which is why the bug was not caught earlier. 

Test run log - https://gist.github.com/mbaijal/8cc74a8e95d3009f96ddc98c1a4c49bd

**Testing:**
- [ ] Unit tests pass
- [ ] e2e tests pass

### DIFF
--- a/tests/e2e/utils/auto-fsx-setup.py
+++ b/tests/e2e/utils/auto-fsx-setup.py
@@ -174,7 +174,7 @@ def create_fsx_iam_service_account():
             "--cluster",
             CLUSTER_NAME,
             "--attach-policy-arn",
-            FSX_IAM_POLICY_NAME,
+            FSX_IAM_POLICY_ARN,
             "--approve",
             "--override-existing-serviceaccounts",
             "--region",
@@ -449,7 +449,7 @@ if __name__ == "__main__":
     CONFIG_FILENAME = args.config_filename
 
     AWS_ACCOUNT_ID = boto3.client("sts").get_caller_identity()["Account"]
-    FSX_IAM_POLICY_NAME = "fsx-csi-driver-policy"
+    FSX_IAM_POLICY_NAME = "fsx-csi-driver-policy"+FSX_FILE_SYSTEM_NAME
     FSX_IAM_POLICY_ARN = f"arn:aws:iam::{AWS_ACCOUNT_ID}:policy/{FSX_IAM_POLICY_NAME}"
     FSX_STATIC_PROVISIONING_STORAGE_CLASS_FILE_PATH = "../../docs/deployment/add-ons/storage/fsx-for-lustre/static-provisioning/sc.yaml"
     FSX_SECURITY_GROUP_ID = ""


goswamig (changes requested):
can  you please upload the test run output?

mbaijal:
> can you please upload the test run output?

I have already linked it above.

goswamig:
```Handling connection for 8080
Handling connection for 8080
INFO     root:_client.py:167 Failed to automatically set namespace.
Traceback (most recent call last):
  File "/Users/mbaijal/.local/lib/python3.8/site-packages/kfp/_client.py", line 163, in __init__
    with open(Client.NAMESPACE_PATH, 'r') as f:
```
What is this? Should we worry about this ?

mbaijal:
> with open(Client.NAMESPACE_PATH, 'r') as f:

@goswamig This is a good point. 
I see this is also happening on the rds-s3 test which uses kfp so this is not related to this or the other fsx/efs PRs. However, it definitely need a dive deep and I have opened an issue here - https://github.com/awslabs/kubeflow-manifests/issues/164

Context based on a very quick initial check - 
It seems to be coming from here - https://kubeflow-pipelines.readthedocs.io/en/stable/_modules/kfp/_client.html and is happening because kfp is not able to set the namespace automatically. 
This comes from the k8s feature defined here - https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#serviceaccount-admission-controller

tests/e2e/utils/auto-fsx-setup.py
```diff
@@ -174,7 +174,7 @@ def create_fsx_iam_service_account():
             "--cluster",
             CLUSTER_NAME,
             "--attach-policy-arn",
-            FSX_IAM_POLICY_NAME,
+            FSX_IAM_POLICY_ARN,
```
goswamig, line 177: how did it work before ?

tests/e2e/utils/auto-fsx-setup.py
```diff
@@ -449,7 +449,7 @@ def footer():
     CONFIG_FILENAME = args.config_filename
 
     AWS_ACCOUNT_ID = boto3.client("sts").get_caller_identity()["Account"]
-    FSX_IAM_POLICY_NAME = "fsx-csi-driver-policy"
+    FSX_IAM_POLICY_NAME = "fsx-csi-driver-policy"+FSX_FILE_SYSTEM_NAME
```
goswamig, line 452: why do we need this?

goswamig (approved):
changes looks fine just few questions.

tests/e2e/utils/auto-fsx-setup.py
mbaijal: This worked because it was running on an existing cluster and since the FSX_IAM_POLICY_NAME was fixed, it already existed. I had an action item to run FSx on a new cluster and thus caught this issue. 
Thus I made `FSX_IAM_POLICY_NAME ` it variable now such that these issues don't slip through.

PR #197, opened 2022-04-21
OIDC for fsx test; cleanup kfp pods

mbaijal:
Resolves #
Removed OIDC set up for FSx by mistake while moving from manual to auto setup. Adding it back
also added it for the dynamic setup of EFS. 

This passed for me because I usually ran all these tests on the same cluster and static efs does setup OIDC. 

**Testing:**
- [ ] Unit tests pass
- [ ] e2e tests pass

### DIFF
--- a/tests/e2e/fixtures/storage_efs_dependencies.py
+++ b/tests/e2e/fixtures/storage_efs_dependencies.py
@@ -275,10 +275,10 @@ def static_provisioning(metadata, region, request, cluster, create_efs_volume):
         efs_claim["claim_name"] = claim_name
 
     def on_delete():
+        kubectl_delete(efs_permissions_filepath)
         kubectl_delete(efs_pvc_filepath)
         kubectl_delete(efs_pv_filepath)
-        kubectl_delete(efs_sc_filepath)
-        kubectl_delete(efs_permissions_filepath)
+        kubectl_delete(efs_sc_filepath)   
 
     return configure_resource_fixture(
         metadata, request, efs_claim, "efs_claim", on_create, on_delete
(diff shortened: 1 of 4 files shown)


tests/e2e/fixtures/storage_efs_dependencies.py
```diff
@@ -287,6 +287,7 @@ def on_delete():
 
 @pytest.fixture(scope="class")
 def dynamic_provisioning(metadata, region, request, cluster):
+    associate_iam_oidc_provider(cluster, region)
```
surajkota, line 290: I think we should just add `--with-oidc` here: https://github.com/awslabs/kubeflow-manifests/blob/main/tests/e2e/fixtures/cluster.py#L22

mbaijal: we are running the test with this fix now, let's change this in next release

surajkota (approved):

rrrkharse (approved):

tests/e2e/fixtures/storage_fsx_dependencies.py
```diff
@@ -117,7 +117,8 @@ def on_delete():
         )
 
         # delete the security group
-        ec2_client.delete_security_group(GroupId=sg_id)
+        # TODO: This needs to be fixed. we need to wait for fsx volume to be deleted but fsx does not provided a deleted status, needs a workaround. 
+        # ec2_client.delete_security_group(GroupId=sg_id)
```
rrrkharse, line 121: can we add a sleep command for now
if we dont delete the group not sure how long till we hit the service limit on num of sgs

mbaijal: I thought of that but I do know a cleaner way to fix this - documented it here - https://github.com/awslabs/kubeflow-manifests/issues/200

Any amount of sleep is not reliable and I think this step does take a while. Since we are not running this test in any pipeline at the moment we should be ok and I will fix this asap post release.

ryansteakley (approved):
Waiting on last half of efs test but rest have passed.

PR #217, opened 2022-05-05
Fix EFS Cleanup to wait for deletion of mount targets

mbaijal:
EFS Cleanup was failing sometimes since there was a delay in the cleanup of mount targets and this caused the `delete-file-system` call to fail. This has been fixed here. 

Testing - 
Tests have passed but with some local hacks - one final clean run pending. 
Logs will be added here - https://gist.github.com/mbaijal/21b1c4f2c6e0de86fc585acaf9a0e571

**Testing:**
- [ ] Unit tests pass
- [ ] e2e tests pass

### DIFF
--- a/tests/e2e/fixtures/storage_efs_dependencies.py
+++ b/tests/e2e/fixtures/storage_efs_dependencies.py
@@ -55,6 +55,17 @@ def wait_on_efs_status(desired_status, efs_client, file_system_id):
 
     wait_for(callback)
 
+def wait_on_mount_target_status(desired_status, efs_client, file_system_id):
+    def callback():
+        response = efs_client.describe_file_systems(
+            FileSystemId=file_system_id,
+        )
+        number_of_mount_targets = response["FileSystems"][0]["NumberOfMountTargets"]
+        print(f"{file_system_id} has {number_of_mount_targets} mount targets .... waiting")
+        if desired_status == "deleted": assert number_of_mount_targets == 0
+        else: assert number_of_mount_targets > 0
+
+    wait_for(callback)
 
 @pytest.fixture(scope="class")
 def install_efs_csi_driver(metadata, region, request, cluster, kustomize):
@@ -96,7 +107,8 @@ def create_efs_driver_sa(
             PolicyName=policy_name,
             PolicyDocument=policy,
         )
-        assert response["Policy"]["Arn"] is not None
+        policy_arn = response["Policy"]["Arn"]
+        assert policy_arn is not None
 
         create_iam_service_account(
             "efs-csi-controller-sa",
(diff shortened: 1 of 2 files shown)


tests/e2e/fixtures/storage_efs_dependencies.py
```diff
@@ -55,6 +55,17 @@ def callback():
...
 
+def wait_on_mount_target_status(desired_status, efs_client, file_system_id):
+    def callback():
+        response = efs_client.describe_file_systems(
+            FileSystemId=file_system_id,
+        )
+        number_of_mount_targets = response["FileSystems"][0]["NumberOfMountTargets"]
+        print(f"{file_system_id} has {number_of_mount_targets} mount targets .... waiting")
+        if desired_status == "deleted": assert number_of_mount_targets == 0
+        else: assert number_of_mount_targets > 0
```
rrrkharse, line 66: is it also necessary to verify desired_status == actual_status?

tests/e2e/fixtures/storage_efs_dependencies.py
```diff
@@ -211,11 +224,14 @@ def on_delete():
...
 
+        wait_on_mount_target_status("deleted", efs_client, fs_id)
+
         # Delete the Filesystem
+
         efs_client.delete_file_system(
             FileSystemId=fs_id,
         )
-        wait_on_efs_status("deleted", efs_client, fs_id)
+        wait_on_efs_status("deleting", efs_client, fs_id)
```
rrrkharse, line 234: why not "deleted" anymore?

tests/e2e/fixtures/storage_efs_dependencies.py
```diff
@@ -358,11 +374,14 @@ def on_delete():
...
```
rrrkharse, line 384: why not "deleted" anymore?

tests/e2e/fixtures/storage_efs_dependencies.py
mbaijal: the deleted mount target does not have a "status" field. I am just using that string to mean that I should be checking for 0 mount targets attached.

tests/e2e/fixtures/storage_efs_dependencies.py
mbaijal: because deleted gets missed since it is a short lived status and soon goes into error state. However I can fix this similar to fsx and wait on the error and make this cleaner. Let me do that.

tests/e2e/fixtures/storage_efs_dependencies.py
mbaijal: Actually the dynamic provisioning here is still missing security group deletion, let me fix that.

mbaijal: done.

tests/e2e/fixtures/storage_efs_dependencies.py
mbaijal: changed to wait on completion of deletion instead same as fsx.

rrrkharse (approved):

ryansteakley (approved):

akartsky (approved):

PR #232, opened 2022-05-18
Change relative path of yamls after moving them to root dir

akartsky:
Change relative path of yamls after moving them to root dir


goswamig (commented):
didnt @rrrkharse  just send another pr wit h same change ?

ryansteakley:
@rrrkharse pr scope was different it was updating the manifests to kubeflow 1.5 upstream. such as `istio 1.9` is now `istio 1.11`

goswamig (approved):

PR #289, opened 2022-07-06
optimize cognito auto-script to look up existed issued certificate

jsitu777:
… region for matching domain before requesting validation

-Optimized script to re-use 'Issued' certificate with matching domain names.

**Testing:**
- [x] Unit tests pass
- [ ] e2e tests pass

### DIFF
--- a/tests/e2e/fixtures/cluster.py
+++ b/tests/e2e/fixtures/cluster.py
@@ -67,6 +67,16 @@ def create_iam_service_account(
     retcode = subprocess.call(cmd)
     assert retcode == 0
 
+
+def get_iam_service_account_roleArn(service_account_name, namespace):
+    cmd = f"kubectl get sa -n {namespace} {service_account_name} -o yaml | yq '.metadata.annotations.\"eks.amazonaws.com/role-arn\"'"
+    #https://docs.python.org/3/library/subprocess.html#subprocess.run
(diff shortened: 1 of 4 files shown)


surajkota:
Hi Jim, thanks for the PR. Although, it is fine to reuse the certificate, for the tutorials purpose lets not make it the default behaviour. The reason is if 2 different people are creating deployments and one of then runs the clean up script, the root certificate will be deleted.
The root certificate is created separately from the subdomain certificate is because customers might have their top level domain outside of route53 as well and in such cases, the script will ask user to do manual verification for root certificate.

What we could do is ask user to supply an existing certificate ARN in the config file if they want to reuse it so they do it consciously but I think we should punt this type of behaviour for using infrastructure as code. With scripting, things can go wrong as we have seen some cases with RDS

Apologize for the miscommunication

tests/e2e/fixtures/cluster.py
```diff
@@ -67,6 +67,16 @@ def create_iam_service_account(
     retcode = subprocess.call(cmd)
     assert retcode == 0
 
+
+def get_iam_service_account_roleArn(service_account_name, namespace):
```
surajkota, line 71: I am fine with this but it might also not be required since the load balancer chart has option of setting service account creation to false

surajkota:
Closing PR to clean up

PR #321, opened 2022-08-10
Bug-Fix: use default manifests in notebook_images test

mbaijal:
The issue here was that the notebook images test was not installing the kubeflow manifests before running the test but while debugging this issue I realized that I do not infact need patched manifests to run this test since we are using the default user namespace and profile. Thus I cleaned up the extra code. 

However, I do still think that the profile_dependencies file should be pulled out from the profile_irsa test to make the code reusable in other tests in the future. 

Test logs: https://gist.github.com/mbaijal/dcaf37afb04cb3f4b7b1760ffb157e08

**Testing:**
- [ ] Unit tests pass
- [ ] e2e tests pass


surajkota (approved):
Can this test be part of test_sanity? I think we have diverged a lot test_profile_irsa, test_cloudwatch, test_prometheous, test_sanity_portforward etc.

Approving for now.

ryansteakley (approved):

PR #324, opened 2022-08-12
Fix efs and fsx auto script OIDC detection. 

mbaijal:
Fix efs and fsx auto script OIDC detection. The existing code was failing to scan the Tags on the providers returned. One issue was that the existing code would scan each Provider in the account and check the tag attached to it. If the tag matched the cluster name we are looking for the method returned a True. However, one edge case this missed was in case some provided were created differently and simply do not have a Tag attached. Though this could be solved easily by putting the code into a try catch block, a better check is to directly query the cluster and see if it has an oidc attached. 
Thus made the change to both the EFS and FSx autoscripts. 

Ran both the EFS and FSx tests and they pass. 

**Testing:**
- [ ] Unit tests pass
- [ ] e2e tests pass

### DIFF
--- a/tests/e2e/utils/auto-efs-setup.py
+++ b/tests/e2e/utils/auto-efs-setup.py
@@ -49,27 +49,11 @@ def verify_oidc_provider_prerequisite():
 
 
 def is_oidc_provider_present() -> bool:
-    iam_client = get_iam_client()
-    oidc_providers = iam_client.list_open_id_connect_providers()[
-        "OpenIDConnectProviderList"
-    ]
-
-    if len(oidc_providers) == 0:
-        return False
-
-    for oidc_provider in oidc_providers:
-        oidc_provider_tags = iam_client.get_open_id_connect_provider(
-            OpenIDConnectProviderArn=oidc_provider["Arn"]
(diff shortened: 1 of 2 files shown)


AlexandreBrown (approved):

ryansteakley (approved):
nice work. One question will the describe_cluster always have oidc and issuer fields ? Can this lead to a key error if they don't exist?

surajkota (approved):

mbaijal:
> nice work. One question will the describe_cluster always have oidc and issuer fields ? Can this lead to a key error if they don't exist?

That is a good question - let me actually try that case out.

mbaijal:
> > nice work. One question will the describe_cluster always have oidc and issuer fields ? Can this lead to a key error if they don't exist?
> 
> That is a good question - let me actually try that case out.

It was hard to replicate the scenario where the cluster did not have an OIDC attached since the test script takes care of it. Additionally, it does not even matter since if the check fails the script just throws an error and does not attempt to fix it.  However, I have put the code into try block such that the exception is raised at the expected point in the calling function.

ryansteakley (approved):

PR #339, opened 2022-08-31
Integrate ACK into kubeflow

akartsky:
Integrate ACK as part of kubeflow


awsconfigs/apps/ack-sagemaker/kustomization.yaml
```diff
@@ -0,0 +1,67 @@
+# cluster scoped ack installation
```
surajkota, line 1: maybe

- [ ] - change dir name to sagemaker-controller
- [ ] - move under common
- [ ] - put everything from upstream under base with a kustomization.yaml
- [ ] - try using kustomize patch for deployment to directly consume from configmap https://github.com/awslabs/kubeflow-manifests/blob/main/awsconfigs/common/aws-alb-ingress-controller/base/deployment_patch.yaml
- [ ] - remove ack.yaml, .DS_store in multiple places
- [ ] - RDS and S3 only deployment bases to consume from new pipeline base

surajkota (changes requested):

awsconfigs/apps/ack-sagemaker/kustomization.yaml
akartsky: Thanks 👍🏻

awsconfigs/common/ack-sagemaker/params.yaml
```diff
@@ -0,0 +1,5 @@
+varReference:
+- path: metadata/annotations
+  kind: ServiceAccount
+- path: spec/template/spec/containers/0/env/1
```
surajkota, line 4: do you still need this? line 4 &5

awsconfigs/common/ack-sagemaker/kustomization.yaml
```diff
@@ -0,0 +1,49 @@
...
+- name: ack-install-parameters
+  envs:
+  - params.env
+generatorOptions:
+  disableNameSuffixHash: true
+vars:
+- name: ACK_SAGEMAKER_OIDC_ROLE
+  objref:
+    kind: ConfigMap
+    name: ack-install-parameters
```
surajkota, line 39: change we change this configmap name to `ack-sagemaker-controller-parameters`?

surajkota: - it would have been good to move this under an overlay named kubeflow
- can we rename `ack-sagemaker` directory name to sagemaker-operator? because that is the official product name
sagemaker-controller is also okay

surajkota (changes requested):
minor comments

surajkota (approved):
lgtm.

Please run the `test_manifests_build` before merging. Also, missing unit tests

PR #358, opened 2022-09-14
Terraform vanilla installation

rrrkharse:
**Description of your changes:**
- Uses terraform to:
  - Create a VPC
  - Create an EKS cluster
  - Install required EKS addons (e.g. vpc_cni, cert_manager, etc.)
  - Installs KF components (e.g. katib, kubeflow pipelines, etc.)
  - Installs ACK sagemaker

- Also uses terraform to delete the above resources

**Testing:**
https://github.com/rrrkharse/kubeflow-manifests/actions/runs/3062922252/jobs/4944430101

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #366, opened 2022-09-15
add subdomain and cognito ingress terraform module

surajkota:
- Module to create subdomain and add test in gh-action
- Module to create ingress. Did not add test since it needs a cluster but have tested it locally. Will add an end to end test with the complete deployment after #358 is merged

**Testing:**
- [x] e2e tests - create and destroy in gh-action

### DIFF
--- a/iaac/terraform/aws-infra/subdomain/outputs.tf
+++ b/iaac/terraform/aws-infra/subdomain/outputs.tf
@@ -1,4 +1,4 @@
-output "subdomain_zone_name" {
+output "aws_route53_subdomain_zone_name" {
   description = "Subdomain which will be used for Kubeflow Platform (e.g. platform.example.com)"
   value       = aws_route53_zone.platform.name
 }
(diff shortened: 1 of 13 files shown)


iaac/terraform/aws-infra/subdomain/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "subdomain_zone_name" {
```
surajkota, line 1: @rrrkharse should this output be also named as `aws_route53_subdomain_zone_name`?

I kept different name to separate from input but it doesnt have to be the case, I think var and outputs can have same names

rrrkharse: var and outputs can have the same name

surajkota: ya, changed it to be consistent

PR #368, opened 2022-09-16
modify makefile to use python installation script for kf install

jsitu777:
Resolves #

-call python installation script in makefile

**Testing:**
- [ ] Unit tests pass
- [x] e2e tests pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,19 +97,12 @@ cleanup-ack-req: verify-cluster-variables
 	yq e '.cluster.region=env(CLUSTER_REGION)' -i tests/e2e/utils/ack_sm_controller_bootstrap/config.yaml
 	cd tests/e2e && PYTHONPATH=.. python3.8 utils/ack_sm_controller_bootstrap/cleanup_sm_controller_req.py
 
-verify-installation-option:
+verify-automation-variables:
 	test $(INSTALLATION_OPTION) || (echo Please export INSTALLATION_OPTION variable; exit 1)
-
-verify-deployment-option:
 	test $(DEPLOYMENT_OPTION) || (echo Please export DEPLOYMENT_OPTION variable; exit 1)
 
-verify-aws-telemetry-option:
-	test $(AWS_TELEMETRY_OPTION) || (echo Please export AWS_TELEMETRY_OPTION variable; exit 1)
+deploy-kubeflow: bootstrap-ack verify-automation-variables
+	cd tests/e2e && PYTHONPATH=.. python3.8 utils/kubeflow_installation.py --$(DEPLOYMENT_OPTION) --installation_option $(INSTALLATION_OPTION)
 
-install-ack: bootstrap-ack
-
-deploy-kf: verify-cluster-variables connect-to-eks-cluster verify-installation-option verify-deployment-option
-	cd tests/e2e && PYTHONPATH=.. python3.8 utils/kubeflow_installation.py --$(DEPLOYMENT_OPTION) --installation_option $(INSTALLATION_OPTION) --aws_telemetry_option $(AWS_TELEMETRY_OPTION) && cd -; 
-
-uninstall-kf: verify-cluster-variables connect-to-eks-cluster verify-installation-option verify-deployment-option verify-cluster-variables
-	cd tests/e2e && PYTHONPATH=.. python3.8 utils/kubeflow_uninstallation.py --$(DEPLOYMENT_OPTION) --installation_option $(INSTALLATION_OPTION) --aws_telemetry_option $(AWS_TELEMETRY_OPTION) && cd -;
+delete-kubeflow: verify-automation-variables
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -97,7 +97,19 @@ cleanup-ack-req: verify-cluster-variables
...
+	test $(INSTALLATION_OPTION) || (echo Please export INSTALLATION_OPTION variable; exit 1)
 
-deploy-kf-vanilla: bootstrap-ack
-	while ! kustomize build deployments/vanilla | kubectl apply -f -; do echo "Retrying to apply resources"; sleep 30; done
-        
+verify-deployment-option:
+	test $(DEPLOYMENT_OPTION) || (echo Please export DEPLOYMENT_OPTION variable; exit 1)
+
+verify-aws-telemetry-option:
+	test $(AWS_TELEMETRY_OPTION) || (echo Please export AWS_TELEMETRY_OPTION variable; exit 1)
```
surajkota, line 107: telemetry has to be set to true by default, so it shouldnt exit if not specified

also lets combine these checks into verify-install-args or something

Makefile
```diff
@@ -97,7 +97,19 @@ cleanup-ack-req: verify-cluster-variables
...
+
+install-ack: bootstrap-ack
+
+deploy-kf: verify-cluster-variables connect-to-eks-cluster verify-installation-option verify-deployment-option
```
surajkota, line 111: bootstrap ack needs to de dependency. ack is installed by default

also should we use install-kubeflow instead of deploy-kf

Makefile
```diff
@@ -97,7 +97,19 @@ cleanup-ack-req: verify-cluster-variables
...
+	cd tests/e2e && PYTHONPATH=.. python3.8 utils/kubeflow_installation.py --$(DEPLOYMENT_OPTION) --installation_option $(INSTALLATION_OPTION) --aws_telemetry_option $(AWS_TELEMETRY_OPTION) && cd -; 
+
+uninstall-kf: verify-cluster-variables connect-to-eks-cluster verify-installation-option verify-deployment-option verify-cluster-variables
```
surajkota, line 114: uninstall-kubeflow

surajkota (changes requested):
Thanks for this PR!
need to add more things so that all deployment options like cognito and other deployment options can be run completely via make but can iterate on that.

Makefile
akartsky: I think
"install" gives an impression it will be installed locally on the client machine 
"deploy" gives an impression that it will be deployed to the cluster 

so i think we should use "deploy-kubeflow"

Makefile
akartsky: ya i think we can combine these three into one :
"verify-installation-option", "verify-deployment-option" and "verify-aws-telemetry-option"
into 
"verify-automation-variables"

Makefile
akartsky: also no need for `&& cd -`
every line runs as a sub-command from the root dir

Makefile
```diff
@@ -97,7 +97,19 @@ cleanup-ack-req: verify-cluster-variables
...
-deploy-kf-vanilla: bootstrap-ack
-	while ! kustomize build deployments/vanilla | kubectl apply -f -; do echo "Retrying to apply resources"; sleep 30; done
-        
+verify-deployment-option:
+	test $(DEPLOYMENT_OPTION) || (echo Please export DEPLOYMENT_OPTION variable; exit 1)
+
+verify-aws-telemetry-option:
+	test $(AWS_TELEMETRY_OPTION) || (echo Please export AWS_TELEMETRY_OPTION variable; exit 1)
+
+install-ack: bootstrap-ack
```
akartsky, line 109: why do we need this ?

Makefile
akartsky: maybe use `delete-kubeflow` instead of `uninstall-kubeflow`

jsitu777:
@akartsky @surajkota adjusted the comments

akartsky (approved):

Makefile
```diff
@@ -97,7 +97,13 @@ cleanup-ack-req: verify-cluster-variables
...
-        
+deploy-kubeflow: bootstrap-ack
+	$(eval DEPLOYMENT_OPTION:=vanilla)
+	$(eval INSTALLATION_OPTION:=kustomize)
+	$(eval AWS_TELEMETRY_OPTION:=enable)
+	cd tests/e2e && PYTHONPATH=.. python3.8 utils/kubeflow_installation.py --deployment_option $(DEPLOYMENT_OPTION) --installation_option $(INSTALLATION_OPTION) --aws_telemetry_option $(AWS_TELEMETRY_OPTION)
+
+delete-kubeflow:
+	$(eval DEPLOYMENT_OPTION:=vanilla)
+	$(eval INSTALLATION_OPTION:=kustomize)
```
surajkota, line 108: nit: doesnt this need telemetry option?

surajkota (approved):
Nicee!!

Makefile
akartsky: No 
`delete-kubeflow` will always delete telemetry

PR #437, opened 2022-09-23
Uninstall ACK last

akartsky:
Uninstall ACK last


surajkota (approved):
PR description is out of date. Problem is when users dont delete CRs created by controller. So it should be uninstalled after user namespace

surajkota (approved):

PR #531, opened 2022-12-19
docsearch integration - main

npepin-hub:
Test of the integration of DocSearch (Search component from Algolia) to extract content from Kubeflow on AWS documentation and make it available to users from the first keystroke.

### DIFF
--- a/website/layouts/partials/hooks/body-end.html
+++ b/website/layouts/partials/hooks/body-end.html
@@ -3,3 +3,25 @@
 <script src="{{ $jsCopy.RelPermalink }}"></script>
 {{ end }}
 
+<script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3"></script>
+
+<script type="text/javascript">
+
+docsearch({
+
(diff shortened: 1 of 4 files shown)


website/layouts/partials/hooks/body-end.html
```diff
@@ -3,3 +3,25 @@
...
 
+<script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3"></script>
+
+<script type="text/javascript">
+
+docsearch({
+
+appId: 'A1FXAERJE4',
+
+apiKey: '88f3a753878f24ffdcd3d786be775545',
```
ryansteakley, line 14: is this apikey vulnerable to being misused by other people if it is being exposed here? Does angolia have some reccomendation for the visibility of this api-key?

npepin-hub: no. This is a public API key that is meant to be used in front end code, It can be regenerated at any time. Used for search queries and to list the indices we have access to.

ryansteakley (approved):

PR #566, opened 2023-02-21
Update Istio version in Development Paths from 1-14 to 1-16

jsitu777:
-Update Istio version in Development Paths from 1-14 to 1-16 for kustomize


deployments/rds-s3/base/kustomization.yaml
```diff
@@ -6,23 +6,23 @@ resources:
...
+  - ../../upstream/common/istio-1-16/istio-namespace/base
+  - ../../upstream/common/istio-1-16/istio-install/base
   # OIDC Authservice
   - ../../../upstream/common/oidc-authservice/base
   # Dex
   - ../../../upstream/common/dex/overlays/istio
   # KNative
   - ../../../upstream/common/knative/knative-serving/overlays/gateways
   - ../../../upstream/common/knative/knative-eventing/base
-  - ../../../upstream/common/istio-1-14/cluster-local-gateway/base
```
ananth102, line 19: Wouldn't this go to `kubeflow-manifests/deployments/` instead of `kubeflow-manifests/`

jsitu777: good catch. updated.

ananth102 (approved):
Nice work!

PR #570, opened 2023-02-23
[Load Balancer] Add target-type Ingress annotation to accommodate updated Istio Service Spec Type: ClusterIP  

jsitu777:
- Newer version of kubeflow upstream uses istio 1-16 which update its service spec type from NodePort to ClusterIP for security purpose (https://github.com/kubeflow/manifests/pull/2357)
- Default target-type was set to `instance` for ingress resource which require service type to be nodeport. To deal with this issue, a target-type annotation was added to set it to be `ip` instead (https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/annotations/#target-type)
 

**Testing:**
- [ ] Unit tests pass
- [x] e2e tests pass


ananth102 (approved):

awsconfigs/common/istio-ingress/overlays/cognito/ingress.yaml
```diff
@@ -7,3 +7,4 @@ metadata:
     alb.ingress.kubernetes.io/auth-idp-cognito: '{"UserPoolArn":"$(CognitoUserPoolArn)","UserPoolClientId":"$(CognitoAppClientId)", "UserPoolDomain":"$(CognitoUserPoolDomain)"}'
     alb.ingress.kubernetes.io/certificate-arn: $(certArn)
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS":443}]'
+    alb.ingress.kubernetes.io/target-type: ip
```
surajkota, line 10: Why not add it to base ingress manifest?

https://github.com/awslabs/kubeflow-manifests/blob/main/awsconfigs/common/istio-ingress/base/ingress.yaml

surajkota (changes requested):

awsconfigs/common/istio-ingress/overlays/cognito/ingress.yaml
jsitu777: moved the annotation into base instead.

surajkota (approved):

PR #572, opened 2023-02-24
[Helm] Update Dex Chart for kubeflow v1.7.0

jsitu777:
- update Dex Helm chart per kubeflow v1.7.0-rc.1 upstream


ryansteakley (commented):
the comments going away is just part of the upstream changes right?

jsitu777:
> the comments going away is just part of the upstream changes right?

Are you referring to the comments in Chart.yaml? The original comments were auto generated when you run `helm create`. We incorporate appVersion and chartVersion in the tool so when it write back to the Chart.yaml the comments go away.

ryansteakley (approved):

PR #595, opened 2023-02-28
[Helm] Update ingress and aws-telemetry Chart for kubeflow v1.7.0

jsitu777:
- update ingress and aws-telemetry helm chart per v1.7.0 kubeflow upstream


charts/common/ingress/cognito/templates/Ingress/istio-ingress-istio-system-Ingress.yaml
```diff
@@ -9,9 +9,9 @@ metadata:
     alb.ingress.kubernetes.io/certificate-arn: '{{ .Values.alb.certArn }}'
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTPS":443}]'
     alb.ingress.kubernetes.io/load-balancer-attributes: routing.http.drop_invalid_header_fields.enabled=true
-    alb.ingress.kubernetes.io/scheme: internet-facing
```
ryansteakley, line 12: Was  this an issue before since it was hard-coded?

jsitu777: We don't want it to be hard-coded, but by default it's internet facing (https://github.com/awslabs/kubeflow-manifests/blob/main/charts/common/ingress/cognito/values.yaml#L7) Some users can configure it to `internal` instead.

ryansteakley (approved):

PR #568, opened 2023-02-21
Update Istio helm charts from kubeflow v1.7 upstream

jsitu777:
**Description of your changes:**

update istio, cluster-local-gateway, and kubeflow-istio-resource helm chart from istio 1-14 to istio 1-16
update installation config and kubeflow installation/uninstallation script to include the above change

**Testing:**
- [ ] Unit tests pass
- [x] e2e tests pass

PR #598, opened 2023-03-02
[Bug] Fix Istio Helm Chart

jsitu777:
- fix istio chart from previous istio-1-14 resources (https://github.com/awslabs/kubeflow-manifests/pull/568)
specifically:
 * remove `stats-filter-1.11` and `tcp-stats-filter-1.11` in Envoy kind
 * remove PodDisruptionBudget per https://github.com/kubeflow/manifests/blob/master/common/istio-1-16/istio-install/base/patches/remove-pdb.yaml
```
Error: INSTALLATION FAILED: envoyfilters.networking.istio.io "stats-filter-1.16" already exists
```

**Testing:**
- [ ] Unit tests pass
- [x] e2e tests pass


ryansteakley:
what was the issue?

ryansteakley (approved):
Ran into same issue, glad its resolved now.

PR #616, opened 2023-03-14
Trainium blog post

krhoyt:
Add Trainium blog post.


surajkota:
I see we have added this blog post before but we need to revert it because it uses a different scheduler which is not installed as part of Kubeflow. Customer might face issues using the instructions. We need to test it

@krhoyt can you please revert from main as well?

PR #830, opened 2024-10-25
Kfp namespace fix

randomseed-42:
Resolves #

New users were not able to use the kubeflow pipelines UI. Graphs and metadata never loaded even though the pieline was correctly compiled and executed. 

I added bash scripts that create a user profile and AWS roles and policies that permit the kfp-pipeline-controller the access to a user profile so that it can automatically edit the namespace, add tags tothe profile, and deploy pipeline pods that enable the use of the pipelines ui via the central dashboard.

**Testing:**
- I have created users and sucessfully ran an eda pipeline on them
- the eda pipeline shows the visualization and metadata in the central dashboard when logged in as the test user


randomseed-42:
I did not mean to create this PR

